PR #5, opened 2018-01-09
Handle ENOSYS when calling `pipe2`

cuviper:
Just because libc has the symbol, doesn't mean the kernel supports it.

Based on rust-lang/rust#42521.


alexcrichton:
Oh dear sorry about that! I guess I copied the code over too soon :(

alexcrichton:
I've published 0.1.9 with this as well if you'd like to send a PR to update rustc/cargo

cuviper:
Thanks!  Will do.

PR #6, opened 2018-02-24
Validate jobserver fd more conservatively

ishitatsuyuki:
Background: I'm planning to write a system-wide jobserver implementation. It uses sockets, not pipe, which will be rejected with the current validation.

Reference implementation in make: https://github.com/mirror/make/blob/48c8a116a914a325a0497721f5d8b58d5bba34d4/src/posixos.c#L124

Also in the docs:

> If your tool determines that the --jobserver-auth option is available in MAKEFLAGS but that the file descriptors specified *are closed*,

So we don't need to force them to be a pipe.


alexcrichton:
Sure!

PR #9, opened 2018-11-20
prepare for a rustc experiment: unsafe before_exec

RalfJung:
See https://github.com/rust-lang/rust/issues/39575 for all the details.

I am not sure if this is what we usually do in such a case. Please let me know if I am doing it wrong.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -563,11 +563,16 @@ mod imp {
             // integers through `string_arg` above.
             let read = self.read.as_raw_fd();
             let write = self.write.as_raw_fd();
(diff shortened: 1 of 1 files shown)


alexcrichton:
Let's continue discussion [on the issue](https://github.com/rust-lang/rust/issues/39575#issuecomment-440329213)

RalfJung:
I am using `[patch]` now so I don't need this any more. Thanks fir the pointer!

```
[patch.crates-io]
jobserver = { git = 'https://github.com/RalfJung/jobserver-rs' }
```

PR #11, opened 2018-11-25
Add windows to travis CI config

andjo403:
as I get a fault during cargo test when running locally and as appveyor did not start for #10 tying this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ rust: stable
 matrix:
   include:
     - os: osx
+    - os: windows
     - rust: stable
     - rust: beta
     - rust: nightly


andjo403:
hmm got the same fault as when I run locally

alexcrichton:
Thanks for the PR! I think there's an error on CI though?

andjo403:
yes I got the error on my computer also but can not find what the problem is. tried to use the rustc version and commit that was used in the latest appveyor run but got the same fault then also.

alexcrichton:
This error may just mean that there's software like `make` installed on AppVeyor and not on Travis? I'd personally prefer to stick to AppVeyor to add CI if necessary

PR #12, opened 2019-01-14
Redox support

jackpot51:


alexcrichton:
Thanks! I'm a little uncomfortable with the amount of ad-hoc duplication here and `#[cfg]` magic, perhaps redox could just have an entirely separate implementation module?

jackpot51:
Sure thing, that would be possible

petrochenkov:
@jackpot51 
Is there still interest in supporting this crate on Redox?
The current implementation is 5 years old, it probably needs an update now.
If there's no interest, then we'll close this PR to cleanup the queue.

jackpot51:
I am fairly certain this is longer needed.

PR #13, opened 2019-05-11
Add support for non-unix and non-windows platforms

bjorn3:
Cross process jobserver and thread cancellation is not supported

This is part of an experiment to port rustc to wasi. (cc https://github.com/rust-lang/miri/issues/722)

cc @RReverser

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1020,7 +1020,6 @@ mod imp {
     pub struct Helper {
         thread: JoinHandle<()>,
         quitting: Arc<AtomicBool>,
-        rx_done: Receiver<()>,
     }
 
     pub fn spawn_helper(client: ::Client,
@@ -1030,39 +1029,29 @@ mod imp {
     {
         let quitting = Arc::new(AtomicBool::new(false));
         let quitting2 = quitting.clone();
-        let (tx_done, rx_done) = mpsc::channel();
         let thread = Builder::new().spawn(move || {
             'outer:
             for () in rx {
                 loop {
                     let res = client.acquire();
-                    if let Err(ref e) = res {
-                        if e.kind() == io::ErrorKind::Interrupted {
-                            if quitting2.load(Ordering::SeqCst) {
-                                break 'outer
-                            } else {
-                                continue
-                            }
-                        }
+                    if quitting2.load(Ordering::SeqCst) {
+                        break 'outer;
                     }
                     f(res);
-                    break
+                    break;
                 }
             }
-            tx_done.send(()).unwrap();
         })?;
 
         Ok(Helper {
             thread: thread,
             quitting: quitting,
-            rx_done: rx_done,
         })
     }
 
     impl Helper {
         pub fn join(self) {
             self.quitting.store(true, Ordering::SeqCst);
-            let _ = self.rx_done.recv();
             drop(self.thread.join());
         }
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1047,3 +1047,10 @@ mod imp {
         }
     }
 }
+
+#[test]
+fn no_helper_deadlock() {
+    let x = crate::Client::new(32).unwrap();
+    let _y = x.clone();
+    std::mem::drop(x.into_helper_thread(|_| {}).unwrap());
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -967,8 +967,7 @@ mod imp {
 #[cfg(not(any(unix, windows)))]
 mod imp {
     use std::io;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -254,6 +254,10 @@ impl Client {
     /// On Unix and Windows this will clobber the `CARGO_MAKEFLAGS` environment
     /// variables for the child process, and on Unix this will also allow the
     /// two file descriptors for this client to be inherited to the child.
+    ///
+    /// This method doesn't exist on platforms other than Unix and Windows, as
+    /// those don't have a real jobserver implementation.
+    #[cfg(any(unix, windows))]
```
RReverser: Isn't it better to leave the method for API compatibility, but make it no-op in the `imp` itself? (like it is in Windows case)

bjorn3: Done

src/lib.rs
```diff
@@ -961,3 +964,121 @@ mod imp {
...
+            rx_done: rx_done,
+        })
+    }
+
+    impl Helper {
+        pub fn join(self) {
+            self.quitting.store(true, Ordering::SeqCst);
+            let dur = Duration::from_millis(10);
+            let mut done = false;
+            for _ in 0..100 {
```
alexcrichton: I don't think this needs the same dance as Unix has with attempting to loop, this can simply set the flag, send a message, and then join the thread and that should work. The Unix code was handling a lot of other weird conditions.

src/lib.rs
```diff
@@ -254,13 +254,16 @@ impl Client {
     /// On Unix and Windows this will clobber the `CARGO_MAKEFLAGS` environment
     /// variables for the child process, and on Unix this will also allow the
     /// two file descriptors for this client to be inherited to the child.
+    ///
+    /// On platforms other than Unix and Windows this does nothing.
```
alexcrichton: It seems like it may be more appropriate to panic on other platforms? It's not supported there so only in-process usage is expected to work

RReverser: That would make impossible to use same code compiled to both native and WASI, without extra `#[cfg(...)]`.

bjorn3: @RReverser it would be possible, as WASI doesn't support spawning processes, so if you were to call `configure` you would likely panic shortly afterwards for spawning a thread anyway. Also because the function keeps existing no `#[cfg(...)]` is required to keep it compiling.

src/lib.rs
```diff
@@ -961,3 +963,100 @@ mod imp {
...
+
+        Ok(Helper {
+            thread: thread,
+            quitting: quitting,
+        })
+    }
+
+    impl Helper {
+        pub fn join(self) {
+            self.quitting.store(true, Ordering::SeqCst);
```
alexcrichton: In addition to setting a quietting flag this will need to send a message on the channel to ensure that the other thread, if asleep, wakes up to see the flag.

src/lib.rs
```diff
@@ -961,3 +963,100 @@ mod imp {
...
+            'outer:
+            for () in rx {
+                loop {
+                    let res = client.acquire();
+                    if quitting2.load(Ordering::SeqCst) {
+                        break 'outer;
+                    }
+                    f(res);
+                    break;
+                }
```
alexcrichton: I think this inner `loop` may not be necessary?

src/lib.rs
bjorn3: `HelperThread` seems to drop the `Receiver`, causing the for loop to stop, which means even setting `quitting` to true is not necessary.

alexcrichton: Ah good point, but due to clones of `jobserver::Client` I'm not sure that's a guarantee that can be relied on?

bjorn3: Every time `Client::into_helper_thread` is called, a new helper thread is created. The `HelperThread` handle is not clonable, so dropping it will cause the only `Sender` to be dropped, causing the thread to stop.

alexcrichton: Can you test out this doesn't block forever then?

```rust
let x = Client::new(32);
let y = x.clone();
x.into_helper_thread().join();
```

bjorn3: It works, do you want me to add a test?

```rust
#[test]
fn no_helper_deadlock() {
    let x = crate::Client::new(32).unwrap();
    let _y = x.clone();
    std::mem::drop(x.into_helper_thread(|_| {}).unwrap());
}
```

alexcrichton: Yeah that'd be great, thanks!

bjorn3: Added it.

alexcrichton:
Thanks!

PR #19, opened 2020-01-10
Support for unique token values

tommilligan:
This PR adds the ability to create a jobserver with known, unique token values with `Client::new_with_unique`, and read those values with `token.data().byte()`.

---

Our use case is synchronising shared resources between multiple tests in our codebase. We create a fifo and populate it with tokens of known value (the bytes `0` to `n`). Each tests must acquire a token before accessing a shared resource - the value of the token then acts as a reference to the resource to use during the test.

For instance, while running tests in parallel, we have `n` test databases set up, but only one test may read/write data at any point.

This allows us to run tests in parallel across multiple threads/languages, independent of any one particular test framework.

---

We've been using this for a while now and it's working well - is this something you're interested in adding to the library? It's not strictly compatible with the `GNU make spec` which states:

> - All tokens are identical.

However, in practice the value of the token is not used anywhere and treated as opaque by all other consumers.

### DIFF
--- a/examples/shared_resources.rs
+++ b/examples/shared_resources.rs
@@ -1,0 +1,69 @@
+//! The jobserver can also be used as a synchronisation primitive for shared
+//! resources.
+//!
+//! By populating the jobserver with unique tokens, we can provide a reference
+//! to a single instance of a shared resource.
+extern crate jobserver;
+
+#[cfg(unix)]
+pub mod imp {
+    use std::sync::mpsc::channel;
+    use std::sync::Arc;
+    use std::thread;
+    use std::time;
+
+    use jobserver::Client;
+
+    fn sleep() {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! I would personally prefer, though, to keep this library to a cross-platform implementation of the jobserver protocol. It may be best, perhaps, to fork this crate and customize it to your needs? The jobserver protocol inherently isn't really that great (it's just simple), so for anything more advanced that doesn't explicitly need integration with `make` itself I'd recommend taking a different strategy.

tommilligan:
Okay, fair enough. Closing.

For anything more complex we'd probably use a shared key/value store, but it's nice for us to have this synchronisation without additional dependencies.

PR #21, opened 2020-01-17
Drop tokens without releasing them

Mark-Simulacrum:
This is useful when dealing with the helper thread interface but wanting to
release the tokens via release_raw at a later point.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -109,6 +109,17 @@ pub struct Client {
 pub struct Acquired {
     client: Arc<imp::Client>,
     data: imp::Acquired,
+    disabled: bool,
+}
+
+impl Acquired {
+    /// This drops the `Acquired` token without releasing the associated token.
+    ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -111,6 +111,29 @@ pub struct Acquired {
...
+        // the fields out and then forget the struct.
+        let (client, data) = unsafe {
+            let Acquired {
+                ref client,
+                ref data,
+            } = &self;
+            let client = std::ptr::read(client);
+            let data = std::ptr::read(data);
+            std::mem::forget(self);
+            (client, data)
```
Mark-Simulacrum: I don't know what a good way to implement this is to be honest :)

I'm happy to also just add a bool to the struct and gate on that in the drop impl if that seems better.

Mark-Simulacrum:
I decided that the unsafe code wasn't worth it and just added a bool to disable the destructor, just seems easier and all around nicer.

alexcrichton:
👍

PR #28, opened 2020-08-29
Fix some clippy warnings

lzutao:
I also remove now unneeded `extern crate` statement since this crate uses 2018 edition.


alexcrichton:
Thanks!

PR #31, opened 2021-04-18
Try blocking reads first to avoid thundering herd problem

### LINKED ISSUE #30: polling pipes instead of blocking reads is inefficient on linux
The approach outlined in this comment

https://github.com/alexcrichton/jobserver-rs/blob/9d5e6da2157a3db4e60e276185292d4f65cdaf0d/src/unix.rs#L118-L133

probably is inefficient on linux. The kernel recently gained an optimization where a write to a pipe only wakes up one reader as long as that one reader empties the pipe: https://github.com/torvalds/linux/commit/0ddad21d3e99c743a3aa473121dc5561679e26bb# Note the `wake_next_reader` flag.

Waking up a polling process will never empty the pipe since it first has to return to userspace before that might issue a read syscall, So it'll end up waking more readers than necessary which leads to lots of unnecessary context switches and wasted CPU cycles.

It would be better to simply attempt reading from the fd and only start polling when it returns `EWOULDBLOCK`.

It might even be beneficial to `dup()` the file descriptor and explicitly set the copy to blocking mode if the current process has no need for non-blocking operation.

Edit: I included a lot of hedge words, but here's the word of god saying that poll-based wakeups indeed do not benefit from that optimization: https://lkml.org/lkml/2020/2/12/1019

the8472:
Linux 5.6 and up optimize the wakeups of pipe readers
when pipes have multiple readers and those readers are blocking.
Always polling defeats that optimization.

https://github.com/torvalds/linux/commit/0ddad21d3e99c743a3aa473121dc5561679e26bb

fixes #30

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -122,10 +122,11 @@ impl Client {
         // fds are set to nonblocking and combined with `pselect`
         // internally.
         //
-        // Here we try to be compatible with both strategies. We
-        // unconditionally expect the file descriptor to be in nonblocking
-        // mode and if it happens to be in blocking mode then most of this
-        // won't end up actually being necessary!
+        // Here we try to be compatible with both strategies. We optimistically
+        // try to read from the file descriptor which then may block, return
+        // a token or indicate that polling is needed.
+        // Blocking reads (if possible) allows the kernel to be more selective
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #35, opened 2021-08-16
Revert "Optionally increase the size of the pipe buffer"

alexcrichton:
This reverts commit 865b7e375a29afbc8283f14729d4663daa858a1e.

Lots of errors about "can't increase jobserver limits" are cropping up in https://github.com/rust-lang/rust/issues/88091, and nothing was really that broken before so let's back this out and figure out a better way to deal with this later.


ehuss (approved):
Thanks!

PR #39, opened 2022-06-29
Optimize `unix::Client::new`: Avoid multi `write`

NobodyXu:

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -29,6 +29,8 @@ impl Client {
         // wrong!
         const BUFFER: [u8; 128] = [b'|'; 128];
 
+        set_nonblocking(client.write.as_raw_fd(), true)?;
+
         while limit > 0 {
             let n = limit.min(BUFFER.len());
 
@@ -63,7 +65,6 @@ impl Client {
                         }
                     }
                     _ => {
-                        set_nonblocking(pipes[1], true)?;
                         return Ok(Client::from_fds(pipes[0], pipes[1]));
                     }
                 }
@@ -73,9 +74,6 @@ impl Client {
         cvt(libc::pipe(pipes.as_mut_ptr()))?;
         drop(set_cloexec(pipes[0], true));
         drop(set_cloexec(pipes[1], true));
-
-        set_nonblocking(pipes[1], true)?;
-
         Ok(Client::from_fds(pipes[0], pipes[1]))
     }
 
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -35,13 +35,11 @@ impl Client {
         // 128 cores.
         const BUFFER: [u8; 128] = [b'|'; 128];
 
-        while limit >= BUFFER.len() {
-            (&client.write).write_all(&BUFFER)?;
-            limit -= BUFFER.len();
-        }
+        while limit > 0 {
+            let n = limit.min(BUFFER.len());
 
-        if limit > 0 {
-            (&client.write).write_all(&BUFFER[..limit])?;
+            (&client.write).write_all(&BUFFER[..n])?;
+            limit -= n;
         }
 
         Ok(client)
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -64,9 +64,7 @@ impl Client {
                             return Err(err);
                         }
                     }
-                    _ => {
-                        return Ok(Client::from_fds(pipes[0], pipes[1]));
-                    }
+                    _ => return Ok(Client::from_fds(pipes[0], pipes[1])),
                 }
             }
         }
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -25,9 +25,8 @@ impl Client {
         let client = unsafe { Client::mk()? };
         // I don't think the character written here matters, but I could be
         // wrong!
-        for _ in 0..limit {
-            (&client.write).write_all(&[b'|'])?;
-        }
+        let v: Vec<u8> = vec![b'|'; limit];
+        (&client.write).write_all(&v)?;
         Ok(client)
     }
 
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -22,7 +22,9 @@ pub struct Acquired {
 
 impl Client {
     pub fn new(mut limit: usize) -> io::Result<Client> {
-        assert!(limit <= 4096, "Can't have limit larger than 4096 since the pipe would block forever when writing to it");
+        if limit > 4096 {
+            return Err(io::Error::new(io::ErrorKind::Other,  "Can't have limit larger than 4096 since the pipe would block forever when writing to it"));
+        }
 
         let client = unsafe { Client::mk()? };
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Instead of performing a heap allocation though could this perhaps reserve a 128-byte buffer as as `static` or `const` and write that instead? I suspect most systems have <128 cores so it'll likely have the same effective effect.

NobodyXu:
> Thanks! Instead of performing a heap allocation though could this perhaps reserve a 128-byte buffer as as `static` or `const` and write that instead? I suspect most systems have <128 cores so it'll likely have the same effective effect.

Done!

NobodyXu:
Oops, I accidentally closed it

alexcrichton:
Oh sorry but what I meant was that instead of having anything heap allocated this could instead still have a loop, but the loop is "chunked" into 128-bytes-at-a-time. So no need for `vec!` , still a need for the loop, and most of the time only one trip through the loop will be taken.

NobodyXu:
> Oh sorry but what I meant was that instead of having anything heap allocated this could instead still have a loop, but the loop is "chunked" into 128-bytes-at-a-time. So no need for `vec!` , still a need for the loop, and most of the time only one trip through the loop will be taken.

Got it, I will do that.

NobodyXu:
> Oh sorry but what I meant was that instead of having anything heap allocated this could instead still have a loop, but the loop is "chunked" into 128-bytes-at-a-time. So no need for `vec!` , still a need for the loop, and most of the time only one trip through the loop will be taken.

Done!

And out of some mysterious reasons, my PR got closed again after I pushed my commits.

src/unix.rs
```diff
@@ -21,13 +21,27 @@ pub struct Acquired {
 }
 
 impl Client {
-    pub fn new(limit: usize) -> io::Result<Client> {
+    pub fn new(mut limit: usize) -> io::Result<Client> {
+        assert!(limit <= 4096, "Can't have limit larger than 4096 since the pipe would block forever when writing to it");
```
alexcrichton, line 25: I think it would be best to not have this be an assert. Other options could include:

* Instead of assert, return an error.
* Perform a non-blocking write and return an error if it returns EWOULDBLOCK.
* Leave the behavior as-is.

I'm wary about having a hardcoded constant here because pipes configuration can vary system-to-system.

src/unix.rs
```diff
@@ -21,13 +21,27 @@ pub struct Acquired {
...
         // I don't think the character written here matters, but I could be
         // wrong!
-        for _ in 0..limit {
-            (&client.write).write_all(&[b'|'])?;
+        //
+        // Also, it is highly unlikely to have a machine with more than
+        // 128 cores.
+        const BUFFER: [u8; 128] = [b'|'; 128];
+
+        while limit >= BUFFER.len() {
```
alexcrichton, line 36: If you'd like this could be tightened up a bit with:

```
while limit > 0 {
    (&client.write).write_all(&BUFFER[..limit.min(BUFFER.len())])?;
    limit -= BUFFER.len();
}
```

src/unix.rs
```diff
@@ -21,13 +21,27 @@ pub struct Acquired {
...
+
         let client = unsafe { Client::mk()? };
+
         // I don't think the character written here matters, but I could be
         // wrong!
-        for _ in 0..limit {
-            (&client.write).write_all(&[b'|'])?;
+        //
+        // Also, it is highly unlikely to have a machine with more than
+        // 128 cores.
```
alexcrichton, line 33: I think it's ok to remove this comment since nowadays I'm actually working on machines with ~100 cores so this "highly unlikely" doesn't seem like it'll be true for very long.

src/unix.rs
NobodyXu: Done!

NobodyXu: I used the first solution since I think that is the most reasonable one.

The second one requires the pipe to be set to non-blocking mode and I prefer not to as I afraid it will break other parts of this crate.

src/unix.rs
NobodyXu: Removed!

src/unix.rs
```diff
@@ -21,13 +21,24 @@ pub struct Acquired {
 }
 
 impl Client {
-    pub fn new(limit: usize) -> io::Result<Client> {
+    pub fn new(mut limit: usize) -> io::Result<Client> {
+        if limit > 4096 {
```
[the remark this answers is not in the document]
NobodyXu: You are right, I should use PIPE_BUF.

According to [this man page](https://man7.org/linux/man-pages/man7/pipe.7.html):

> POSIX.1 says that writes of less than PIPE_BUF bytes must be
       atomic: the output data is written to the pipe as a contiguous
       sequence.  Writes of more than PIPE_BUF bytes may be nonatomic:
       the kernel may interleave the data with data written by other
       processes.  POSIX.1 requires PIPE_BUF to be at least 512 bytes.
       (On Linux, PIPE_BUF is 4096 bytes.)  The precise semantics depend
       on whether the file descriptor is nonblocking (O_NONBLOCK),
       whether there are multiple writers to the pipe, and on n, the
       number of bytes to be written:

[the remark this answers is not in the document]
NobodyXu: Done!

src/unix.rs
```diff
@@ -21,13 +22,24 @@ pub struct Acquired {
 }
 
 impl Client {
-    pub fn new(limit: usize) -> io::Result<Client> {
+    pub fn new(mut limit: usize) -> io::Result<Client> {
+        if limit > PIPE_BUF {
+            return Err(io::Error::new(io::ErrorKind::Other,  "Can't have limit larger than 4096 since the pipe would block forever when writing to it"));
+        }
```
alexcrichton, line 28: Personally I still don't think this should be done. I'm pretty sure that the limit of a pipe is a runtime-configurable thing in the kernel and not a static constant someone can write down in libc.

NobodyXu: > I'm pretty sure that the limit of a pipe is a runtime-configurable thing in the kernel and not a static constant someone can write down in libc.

You are right, this `libc::PIPE_BUF` only specifies the minimum size of pipe.

I would switch non-blocking then, but I would have to remove `O_NONBLOCK` before returning from `new`.

NobodyXu: Done.

NobodyXu:
> Can the `set_nonblocking` all happen within `Client::new` with comments as to what it's doing? Otherwise having it spread out across a few methods makes it harder to figure out what's going on.

Sure!

I was thinking about adding `O_NONBLOCK` to `pipe2`, that's why I added it to `mk.

NobodyXu:
> Can the `set_nonblocking` all happen within `Client::new` with comments as to what it's doing? Otherwise having it spread out across a few methods makes it harder to figure out what's going on.

Done!

PR #40, opened 2022-07-01
Fix potentially dropped `Client` in `configure`

### LINKED ISSUE #25: Client::configure is unsafe
If you do something like:
```
if let Some(client) = unsafe { jobserver::Client::from_env() } {
    client.configure(&mut cmd);
}
```

Where `cmd` is an existing `Command`. 

By the time you execute the command and the `pre_cmd` that `Client::configure` set up, the `Client` instance has already been dropped. And dropping the `Client` actually closes the file descriptors (which is documented in `Client::from_env`. So by the time the pre-command executes, the file descriptors are either closed, or worse, were opened for something else. Which reminds me of an old issue we had with cargo a long time ago.

It seems like `Client::configure` should have an explicit requirement on the lifetime of the command not exceeding that of the client.

NobodyXu:
Fixed #25 

### DIFF
--- a/tests/client-dropped-before-command.rs
+++ b/tests/client-dropped-before-command.rs
@@ -18,6 +18,7 @@ fn server() {
     let mut cmd = Command::new(me);
     cmd.env("I_AM_THE_CLIENT", "1");
     client.configure(&mut cmd);
+    drop(client);
 
     let Output {
         status,
--- a/tests/client-dropped-before-command.rs
+++ b/tests/client-dropped-before-command.rs
@@ -26,7 +26,7 @@ fn server() {
         stderr,
     } = cmd.output().unwrap();
 
-    assert!(status.success());
+    assert!(status.success(), "{:#?}", String::from_utf8_lossy(&stderr));
     assert_eq!(&*stderr, b"hello!\n");
 }
 
--- a/tests/client-dropped-before-command.rs
+++ b/tests/client-dropped-before-command.rs
@@ -1,0 +1,37 @@
+use std::env;
+use std::process::{Command, Output};
+
+use jobserver::Client;
+
+fn main() {
+    if env::var("I_AM_THE_CLIENT").is_ok() {
+        client();
+    } else {
+        server();
+    }
+}
+
+fn server() {
+    let me = env::current_exe().unwrap();
+    let client = Client::new(1).unwrap();
+
+    let mut cmd = Command::new(me);
+    cmd.env("I_AM_THE_CLIENT", "1");
+    client.configure(&mut cmd);
+
+    let Output {
+        status,
+        stdout: _stdout,
+        stderr,
+    } = cmd.output().unwrap();
+
+    assert!(status.success());
+    assert_eq!(&*stderr, b"hello!");
+}
+
+fn client() {
+    let client = unsafe { Client::from_env().unwrap() };
+    let acq = client.acquire().unwrap();
+    eprintln!("hello!");
+    drop(acq);
+}
--- a/tests/client-dropped-before-command.rs
+++ b/tests/client-dropped-before-command.rs
@@ -17,14 +17,14 @@ fn server() {
 
     let mut cmd = Command::new(me);
     cmd.env("I_AM_THE_CLIENT", "1");
-    client.configure(&mut cmd);
-    drop(client);
 
     let Output {
         status,
         stdout: _stdout,
         stderr,
-    } = cmd.output().unwrap();
+    } = client
+        .configure_and_run(&mut cmd, |cmd| cmd.output())
+        .unwrap();
 
     assert!(status.success(), "{:#?}", String::from_utf8_lossy(&stderr));
     assert_eq!(&*stderr, b"hello!\n");
--- a/tests/client-dropped-before-command.rs
+++ b/tests/client-dropped-before-command.rs
@@ -27,7 +27,7 @@ fn server() {
     } = cmd.output().unwrap();
 
     assert!(status.success());
-    assert_eq!(&*stderr, b"hello!");
+    assert_eq!(&*stderr, b"hello!\n");
 }
 
 fn client() {
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Install Rust (rustup)
       run: rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
       shell: bash
-    - run: cargo test
+    - run: cargo test --all-features
 
   rustfmt:
     name: Rustfmt
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,9 @@ An implementation of the GNU make jobserver for Rust
 """
 edition = "2018"
 
+[dependencies]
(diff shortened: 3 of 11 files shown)


NobodyXu:
> On second read, I don't know whether this is safe. If this attempts to destroy the `Arc<T>` closure in the `pre_exec` closure I don't think that this is safe.

Oops, I didn't think of that.

I would probably revert it and fix that with lifetime subtype/variance.

NobodyXu:
> On second read, I don't know whether this is safe. If this attempts to destroy the `Arc<T>` closure in the `pre_exec` closure I don't think that this is safe.

I've fixed that by using lifetime subtyping to ensure that `&self` always outlives `&Command`.

alexcrichton:
I think a better solution might be to use some sort of leaking method in the forked process. I'm wary to change the public API here and I'm not actually sure that this fixes the original issue (while it probably fixes the one specific example I'm not sure that this fixes all usages).

NobodyXu:
> I think a better solution might be to use some sort of leaking method in the forked process. I'm wary to change the public API here and I'm not actually sure that this fixes the original issue (while it probably fixes the one specific example I'm not sure that this fixes all usages).

That sounds like a good idea.

I will do it tomorrow.

NobodyXu:
@alexcrichton I've fixed this by leaking the `Arc`.

alexcrichton:
Reading over this, can you confirm the behavior one way or another? This now actually seems like it'll legitimately leak the memory in the host process.

alexcrichton:
I'm more worried about the `Command` in the original parent process where the `Arc` is leaked I believe.

NobodyXu:
> I'm more worried about the `Command` in the original parent process where the `Arc` is leaked I believe.

Thanks for pointing out!

Turns out that I missed that one :D

NobodyXu:
> Looks good to me, can you expand in the comment why the arc is leaked though?

[`Arc::increment_strong_count`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.increment_strong_count) increments the strong count of the `Arc` by one, thus it is leaked unless we pair it with `Arc::decrement_strong_count`.

NobodyXu:
> what I would like the code to document is _why_ it's being leaked because in the abstract that's a pretty weird thing to do intentionally.

Hmmm, `Arc::increment_strong_count(p.0)` is equivalent to the following code:

```rust
let arc = mem::ManuallyDrop::new(unsafe { Arc::from_raw(p.0) });
let _arc_clone: mem::ManuallyDrop<_> = arc.clone();
```

NobodyXu:
> Sorry yes I understand that this is being leaked, what I would like the code to document is _why_ it's being leaked because in the abstract that's a pretty weird thing to do intentionally.

Basically, `Arc` uses reference counting to keep track of the references and it is deallocated when the reference counting reaches `0`.

Using `Arc::increment_strong_count` increment the reference counting by 1 without creating an `Arc`, thus prevents the reference counting from ever reaching 0 and being dropped.

alexcrichton:
Sorry I think more things need to be added here:

* There should be a test from the original issue to assert that it works.
* There should ideally be a test that nothing is leaked in the host process. Currently I think a leak happens.
* It is not necessary to document why incrementing the strong count leaks the `Arc`. What I think is necessary is to document why anything is being leaked in the first case.
* Again though I think it would be good to confirm that the child process actually runs the destructor of the `Arc`. If it doesn't do that then there's no need to deal with leaking anyway.

NobodyXu:
@alexcrichton I've checked the source code of [`Command::do_exec`](https://github.com/rust-lang/rust/blob/1cd72b734318720adb99dc72147bb8169ef76ffe/library/std/src/sys/unix/process/process_unix.rs#L358), it does not destructs the `pre_exec` function we pass to them.

However, the use of `pre_exec` itself prevents the use of `posix_spawn`, which can use `vfork` to avoid duplicating the virtual memory mapping, so I think we should avoid the use of `pre_exec` by creating the pipe without `O_CLOEXEC` flag.

NobodyXu:
@alexcrichton I've added a test case to assert it works.

tests/client-dropped-before-command.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+fn server() {
+    let me = env::current_exe().unwrap();
+    let client = Client::new(1).unwrap();
+
+    let mut cmd = Command::new(me);
+    cmd.env("I_AM_THE_CLIENT", "1");
+    client.configure(&mut cmd);
```
alexcrichton, line 20: To test the original bug the `client` needs to be dropped after this, otherwise I think this test already passes on `mian`

alexcrichton (approved):

alexcrichton:
The test is also not currently running because it is `#[test]`

NobodyXu:
> The test is also not currently running because it is not `#[test]`

Since I modeled the tests after `client-of-myself.rs`, I just need to add `harness = false` to the `Cargo.toml`

I don't think `#[test]` would work here as we manually implement `main`.

NobodyXu:
@alexcrichton The test `client-dropped-before-command` failed on windows, however I have no idea how to fix it.
Can we merge this PR and then fix this in another PR?

alexcrichton:
No I would prefer to not land a broken test that doesn't succeed on all platforms.

NobodyXu:
> No I would prefer to not land a broken test that doesn't succeed on all platforms.

Fair enough.

After looking at the mod `windows`, I think the only way to fix it is to rework the API.
I would redesign the API then.

NobodyXu:
@alexcrichton I figured it out, there actually isn't any problem with the my implementation of the API.

> No I would prefer to not land a broken test that doesn't succeed on all platforms.

We actually do not support sharing the jobserver across processes on windows, but I did not disable this test for windows.

NobodyXu:
Ignore the previous comment, it was actually wrong.

I mistook `wasm.rs` for `windows.rs`...

NobodyXu:
@alexcrichton I have fixed the CI failure, can you review this again please?

The latest commits fixed this by changing `configure` to `configure_and_run`, taking `Command` by value and requires a function `f` that performs the spawning.

`configure_and_run` requires user to perform spawning in `f` to ensure that `Client` must be alive at when spawning and once the function exits, it will clear any environment variable it sets.

This also avoids the use of `std::os::unix::process::CommandExt` to register `pre_exec`, meaning that on unix, `Command::spawn` can use `posix_spawn` to efficiently spawn a new process using `vfork`.

I also added support for `tokio::process::Command` behind a feature flag `tokio` so now people using `tokio` can also use this crate trivally without manually converting between `std::process::Command` and `tokio::process::Command`.

alexcrichton:
Sorry this has grown to the point that I don't really know what's going on internally any more. If it's this difficult to fix the original issue then it may not be worth fixing or otherwise could try to just result in a better error or something like that.

alexcrichton:
Yes I understand it's the API itself that's causing the issue. This is so significantly different from before that I don't want to pull it into this crate. If you'd like though you could maintain a separate crate which depends on this one.

NobodyXu:
> Yes I understand it's the API itself that's causing the issue. This is so significantly different from before that I don't want to pull it into this crate. If you'd like though you could maintain a separate crate which depends on this one.

Understandable, I might just fork this crate and create jobserver2.

PR #42, opened 2022-07-01
Add `configure_make`: Set `MAKEFLAGS` & `MFLAGS`

### LINKED ISSUE #24: Client::configure can't be used to call e.g. make
`Client::configure` only sets `CARGO_MAKEFLAGS`, and there is no way to retrieve the value it sets in the `Command` because there is no API for that. So it's not possible to set `MAKEFLAGS` yourself to pass cargo's jobserver to GNU make in a build script.

NobodyXu:
Fixed #24 

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -290,15 +290,41 @@ impl Client {
     ///
     /// On platforms other than Unix and Windows this panics.
     pub fn configure(&self, cmd: &mut Command) {
-        let arg = self.inner.string_arg();
-        // Older implementations of make use `--jobserver-fds` and newer
-        // implementations use `--jobserver-auth`, pass both to try to catch
-        // both implementations.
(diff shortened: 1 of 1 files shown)


alexcrichton:
I don't want to undo https://github.com/alexcrichton/jobserver-rs/commit/6ac9e7ca16fd792865cf16f6d2c8155dfbf299f3, but I think making this opt-in is ok.

NobodyXu:
> I don't want to undo [6ac9e7c](https://github.com/alexcrichton/jobserver-rs/commit/6ac9e7ca16fd792865cf16f6d2c8155dfbf299f3), but I think making this opt-in is ok.

I've added a new function `configure_make` that sets `MAKEFLAGS` and `MFLAGS` in additional to `CARGO_MAKEFLAGS` to keep backward compatibility.

alexcrichton:
👍

PR #49, opened 2023-02-25
Support `--jobserver-auth=fifo:PATH`

### LINKED ISSUE #47: New jobserver type in GNU Make 4.4
https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html#POSIX-Jobserver

On systems that support it GNU Make 4.4 will now use `--jobserver-auth=fifo:PATH` where `PATH` points to a named pipe rather than ` --jobserver-auth=R,W`. I haven't checked if this breaks anything, but based on looking at the code it seems like it will cause the jobserver to be ignored entirely.

weihanglo:
GNU `make` 4.4, released in October 2022[^1]. The jobserver defaults
to use named pipes (via `mkfifo(3)`) on supported platforms by
introducing a new IPC style `--jobserver-auth=fifo:PATH`, which `PATH`
is the path of fifo[^2].

This commit makes sure that the new style `--jobserver-auth=fifo:PATH`
can be forwarded to inherited processes correctly.

The support of creating a new client with named pipe will come as a
follow-up pull request.

[^1]: https://lists.gnu.org/archive/html/info-gnu/2022-10/msg00008.html
[^2]: https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -1,10 +1,11 @@
 use libc::c_int;
 
-use std::fs::File;
+use std::fs::{File, OpenOptions};
 use std::io::{self, Read, Write};
 use std::mem;
 use std::mem::MaybeUninit;
 use std::os::unix::prelude::*;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::ptr;
 use std::sync::{Arc, Once};
@@ -12,9 +13,11 @@ use std::thread::{self, Builder, JoinHandle};
 use std::time::Duration;
 
 #[derive(Debug)]
-pub struct Client {
-    read: File,
-    write: File,
+pub enum Client {
+    /// `--jobserver-auth=R,W`
+    Pipe { read: File, write: File },
+    /// `--jobserver-auth=fifo:PATH`
+    Fifo { file: File, path: PathBuf },
 }
 
 #[derive(Debug)]
@@ -30,16 +33,18 @@ impl Client {
         // wrong!
         const BUFFER: [u8; 128] = [b'|'; 128];
 
-        set_nonblocking(client.write.as_raw_fd(), true)?;
+        let mut write = client.write();
+
+        set_nonblocking(write.as_raw_fd(), true)?;
 
         while limit > 0 {
             let n = limit.min(BUFFER.len());
 
-            (&client.write).write_all(&BUFFER[..n])?;
+            write.write_all(&BUFFER[..n])?;
             limit -= n;
         }
 
-        set_nonblocking(client.write.as_raw_fd(), false)?;
+        set_nonblocking(write.as_raw_fd(), false)?;
 
         Ok(client)
     }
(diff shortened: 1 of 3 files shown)


kaspar030:
Thanks for working on this! I'm looking forward to the follow-up PR, any chance to (draft) PR that already?

weihanglo:
Haven't yet prepared the PR. Off the top of my head, make `fn new()` defaults to fifo is not a solution in the near feature, as build script in Cargo (which is a big dependant of `jobserver`) could still have a chance to invoke old `make`. Haven't thought about how to deal with breaking changes though.

I'd like to hear what do you expect :)

src/unix.rs
```diff
@@ -77,6 +82,32 @@ impl Client {
...
+        }
+        let path = match parts.next() {
+            Some(p) => Path::new(p),
+            None => return None,
+        };
+        let file = match OpenOptions::new().read(true).write(true).open(path) {
+            Ok(f) => f,
+            Err(_) => return None,
+        };
+        drop(set_cloexec(file.as_raw_fd(), true));
```
alexcrichton, line 102: I think this can be skipped since `OpenOptions` should always do this for us?

src/unix.rs
```diff
@@ -216,12 +267,19 @@ impl Client {
...
+        let is_pipe = match self {
+            Client::Pipe { .. } => true,
+            _ => false,
+        };
         unsafe {
             cmd.pre_exec(move || {
                 set_cloexec(read, false)?;
-                set_cloexec(write, false)?;
+                // Only simple pipe style needs to configure both ends.
+                if is_pipe {
```
alexcrichton, line 280: I think that this `if` can be used to skip the entire `pre_exec` perhaps?

src/unix.rs
weihanglo: You're absolutely right. [std does do that](https://github.com/rust-lang/rust/blob/49b9cc5139dd4d11ef78dc08c1f9170de5b1ca39/library/std/src/sys/unix/fs.rs#L995-L996).

weihanglo:
Hi Alex. I amend the first commit to include your suggestions. Also add a new commit to test GNU Make 4.4 in CI, it compiles make from source as a temporary workaround.

alexcrichton (approved):

weihanglo:
Thanks for the quick review, @alexcrichton! Sorry for bombarding your inbox again. If you've got time, it would be helpful if we have a release for this update. No hurry and take your time :)

src/unix.rs
```diff
@@ -202,22 +249,31 @@ impl Client {
     }
 
     pub fn string_arg(&self) -> String {
-        format!("{},{}", self.read.as_raw_fd(), self.write.as_raw_fd())
+        match self {
+            Client::Pipe { read, write } => format!("{},{}", read.as_raw_fd(), write.as_raw_fd()),
+            Client::Fifo { path, .. } => format!("fifo:{}", path.to_str().unwrap()),
```
NobodyXu, line 254: I think it's a bad idea to use the new `fifo:` syntax by default since this is a new syntax and not every program supports this.

weihanglo: Correct. That's why we only support this syntax when inheriting from the environment, and it's really in a fifo style.

NobodyXu: > Correct. That's why we only support this syntax when inheriting from the environment, and it's really in a fifo style.

Yeah this makes sense, though IMO it would be better to keep using the old syntax.

I do appreciate this new syntax since it means we no longer have to register a callback to `Command::pre_exec`, which blocks the use of `vfork`.

Issue #51, opened 2023-03-10
Improve diagnostics on closed or corrupted file descriptors


petrochenkov:
This issue is a continuation of https://github.com/alexcrichton/jobserver-rs/issues/27#issuecomment-701020994.

Jobserver users may want to report warnings or errors if the jobserver cannot be correctly inherited from the environment.

For example, rustc previously ICEd on jobserver errors (https://github.com/rust-lang/rust/issues/46981), but now it either swallows them (when jobserver acquire or release fails) or silently creates a new jobserver.
However, for diagnosing issues in the build environment we'd ideally want to
- Warn if the jobserver descriptors were passed but couldn't be parsed or turned out closed
  - both name of the used env var (e.g `MAKEFLAGS`) and possible parsing errors should be available for reporting
- Error or warn if the jobserver descriptors are open/accessible, but corrupted, i.e. reopened by someone else and no longer refer to pipes
  - name of the used env var should be available in this case as well

Cargo probably wants something like this as well.

This crate can support producing extended error types from `Client::from_env` to support such diagnostics.

petrochenkov:
This was fixed in the PRs linked above, I forgot to close.

PR #52, opened 2023-03-12
Diagnostic: extend from_env function

BelovDV:
Implements major part of #51.

This pr adds error type for `from_env()` and returns possibility to check if fds refer to pipe as optional feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,9 @@ edition = "2018"
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.50"
 
+[features]
+check_pipe = []
+
 [dev-dependencies]
 futures = "0.1"
 num_cpus = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,6 @@ edition = "2018"
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.50"
 
-[features]
-do_not_check_pipe = []
-
 [dev-dependencies]
 futures = "0.1"
 num_cpus = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 libc = "0.2.50"
 
 [features]
-check_pipe = []
+do_not_check_pipe = []
 
 [dev-dependencies]
 futures = "0.1"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,8 +29,8 @@
 //!
 //! // See API documentation for why this is `unsafe`
 //! let client = match unsafe { Client::from_env() } {
-//!     Ok(client) => client,
-//!     Err(_) => panic!("client not configured"),
+//!     Some(client) => client,
+//!     None => panic!("client not configured"),
 //! };
 //! ```
 //!
@@ -269,7 +269,7 @@ impl Client {
     ///
     /// Note, though, that on Windows it should be safe to call this function
     /// any number of times.
-    pub unsafe fn from_env() -> Result<Client, ErrFromEnv> {
+    pub unsafe fn from_env_ext() -> Result<Client, ErrFromEnv> {
         let (env, var) = ["CARGO_MAKEFLAGS", "MAKEFLAGS", "MFLAGS"]
             .iter()
             .map(|&env| env::var(env).map(|var| (env, var)))
@@ -287,6 +287,14 @@ impl Client {
             Ok(c) => Ok(Client { inner: Arc::new(c) }),
             Err(err) => Err(ErrFromEnv::PlatformSpecific { err, env, var }),
         }
+    }
+
+    /// Attempts to connect to the jobserver specified in this process's
+    /// environment.
+    ///
+    /// Wraps `from_env_ext` and discards error details.
+    pub unsafe fn from_env() -> Option<Client> {
+        Self::from_env_ext().ok()
     }
 
     /// Acquires a token from this jobserver client.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -250,8 +250,8 @@ impl Client {
     /// with `CLOEXEC` so they're not automatically inherited by spawned
     /// children.
     ///
-    /// There is a feature to configure, will this function check if provided
-    /// files are actually pipes.
+    /// On unix if `unix_check_is_pipe` enabled this function will check if
+    /// provided files are actually pipes.
     ///
     /// # Safety
     ///
@@ -269,7 +269,7 @@ impl Client {
     ///
     /// Note, though, that on Windows it should be safe to call this function
     /// any number of times.
-    pub unsafe fn from_env_ext() -> Result<Client, ErrFromEnv> {
+    pub unsafe fn from_env_ext(check_pipe: bool) -> Result<Client, ErrFromEnv> {
         let (env, var) = ["CARGO_MAKEFLAGS", "MAKEFLAGS", "MFLAGS"]
             .iter()
             .map(|&env| env::var(env).map(|var| (env, var)))
@@ -283,7 +283,11 @@ impl Client {
             .ok_or(ErrFromEnv::IsNotConfigured)?;
 
         let s = var[pos + arg.len()..].split(' ').next().unwrap();
-        match imp::Client::open(s) {
+        #[cfg(unix)]
+        let imp_client = imp::Client::open(s, check_pipe);
+        #[cfg(not(unix))]
+        let imp_client = imp::Client::open(s);
+        match imp_client {
             Ok(c) => Ok(Client { inner: Arc::new(c) }),
             Err(err) => Err(ErrFromEnv::PlatformSpecific { err, env, var }),
         }
@@ -294,7 +298,7 @@ impl Client {
     ///
     /// Wraps `from_env_ext` and discards error details.
     pub unsafe fn from_env() -> Option<Client> {
-        Self::from_env_ext().ok()
+        Self::from_env_ext(false).ok()
     }
 
     /// Acquires a token from this jobserver client.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,7 +159,7 @@ pub enum ErrFromEnv {
     /// Cannot connect following this process's environment.
     PlatformSpecific {
         /// Error.
-        err: imp::ErrFromEnv,
+        err: io::Error,
         /// Name of gotten env var.
         env: &'static str,
         /// Value of gotten env var.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -167,6 +167,28 @@ pub enum ErrFromEnv {
     },
 }
 
+impl std::fmt::Display for ErrFromEnv {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ErrFromEnv::IsNotConfigured => {
+                write!(f, "couldn't find relevant environment variable")
+            }
+            ErrFromEnv::PlatformSpecific { err, env, var } => {
+                write!(f, "{err} ({env}={var}")
+            }
+        }
+    }
+}
+
+impl std::error::Error for ErrFromEnv {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            ErrFromEnv::IsNotConfigured => None,
+            ErrFromEnv::PlatformSpecific { err, .. } => Some(err),
+        }
+    }
+}
+
 impl Client {
     /// Creates a new jobserver initialized with the given parallelism limit.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,8 +29,8 @@
 //!
 //! // See API documentation for why this is `unsafe`
 //! let client = match unsafe { Client::from_env() } {
-//!     Some(client) => client,
-//!     None => panic!("client not configured"),
+//!     Ok(client) => client,
+//!     Err(_) => panic!("client not configured"),
 //! };
 //! ```
 //!
@@ -151,6 +151,22 @@ struct HelperInner {
     consumer_done: bool,
 }
 
+/// Error type for `from_env` function.
+#[derive(Debug)]
+pub enum ErrFromEnv {
+    /// There isn't env var, that describes jobserver to inherit.
+    IsNotConfigured,
+    /// Cannot connect following this process's environment.
+    PlatformSpecific {
+        /// Error.
+        err: imp::ErrFromEnv,
+        /// Name of gotten env var.
+        env: &'static str,
+        /// Value of gotten env var.
+        var: String,
+    },
+}
+
 impl Client {
     /// Creates a new jobserver initialized with the given parallelism limit.
     ///
@@ -197,8 +213,9 @@ impl Client {
     /// # Return value
     ///
     /// If a jobserver was found in the environment and it looks correct then
-    /// `Some` of the connected client will be returned. If no jobserver was
-    /// found then `None` will be returned.
+    /// `Ok` of the connected client will be returned. If no relevant env var
+    /// was found then `Err(IsNotConfigured)` will be returned. In other cases
+    /// `Err(PlatformSpecific)` will be returned.
     ///
     /// Note that on Unix the `Client` returned **takes ownership of the file
     /// descriptors specified in the environment**. Jobservers on Unix are
@@ -210,6 +227,9 @@ impl Client {
     /// Additionally on Unix this function will configure the file descriptors
     /// with `CLOEXEC` so they're not automatically inherited by spawned
     /// children.
+    ///
+    /// There is a feature to configure, will this function check if provided
+    /// files are actually pipes.
     ///
     /// # Safety
     ///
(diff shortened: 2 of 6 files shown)


src/unix.rs
```diff
@@ -81,61 +88,66 @@ impl Client {
...
+        match (Self::from_fifo(s), Self::from_pipe(s)) {
+            (Some(Ok(c)), _) | (_, Some(Ok(c))) => Ok(c),
+            (Some(Err(e)), _) | (_, Some(Err(e))) => Err(e),
+            (None, None) => Err(ErrFromEnv::ParseEnvVar),
+        }
     }
 
     /// `--jobserver-auth=fifo:PATH`
-    fn from_fifo(s: &str) -> Option<Client> {
+    fn from_fifo(s: &str) -> Option<Result<Client, ErrFromEnv>> {
```
weihanglo, line 100: Would it be more intuitive with `Result<Option<Client>, ErrFromEnv>`? So that we can return `Err(…)` instead of `Some(Err(…))`.

Or even more, just `Result<Client, ErrFromEnv>` and return `Err(ParseEnvVar)` for the old `None` code path.

src/unix.rs
```diff
@@ -81,61 +88,66 @@ impl Client {
         Ok(Client::from_fds(pipes[0], pipes[1]))
     }
 
-    pub unsafe fn open(s: &str) -> Option<Client> {
-        Client::from_fifo(s).or_else(|| Client::from_pipe(s))
+    pub unsafe fn open(s: &str) -> Result<Client, ErrFromEnv> {
+        match (Self::from_fifo(s), Self::from_pipe(s)) {
```
weihanglo, line 92: This is a bit unfortunate. Previously if people is using fifo-style jobserver, they don't need to read from pipe. Now they need both.

src/unix.rs
```diff
@@ -81,61 +88,66 @@ impl Client {
...
+        Some(Ok(Client::Fifo {
             file,
             path: path.into(),
-        })
+        }))
     }
 
     /// `--jobserver-auth=R,W`
-    unsafe fn from_pipe(s: &str) -> Option<Client> {
+    unsafe fn from_pipe(s: &str) -> Option<Result<Client, ErrFromEnv>> {
```
weihanglo, line 120: ditto

src/unix.rs
```diff
@@ -385,8 +397,26 @@ impl Helper {
...
 }
 
-fn is_valid_fd(fd: c_int) -> bool {
-    unsafe { libc::fcntl(fd, libc::F_GETFD) != -1 }
+fn check_fd(fd: c_int) -> bool {
+    #[cfg(feature = "check_pipe")]
+    unsafe {
+        let mut stat = mem::zeroed();
+        if libc::fstat(fd, &mut stat) == 0 {
+            // On android arm and i686 mode_t is u16 and st_mode is u32,
```
weihanglo, line 405: Could you provide a lint to the reference documentation or code? Thank you!

weihanglo (commented):
This looks good to me in general. Let's see other's thought on the change :)

src/unix.rs
BelovDV: Current `None` isn't actual error. Wouldn't it be less intuitive to use `Err` in this case? I'm not sure.

As for `Result<Option<Client>, ErrFromEnv>` - it's better, than my variant, if `Option` will remain here - I'll change to this.

src/unix.rs
BelovDV: I'll check, if both `fifo:PATH` and `R,W` can be provided, but even if they can and will. Both clients will be created, but only one `from_fifo` will be used. Is smth bad in calling `from_pipe` after succeeded `from_fifo`?

src/unix.rs
weihanglo: I proposed removing `Option` because [there](https://github.com/alexcrichton/jobserver-rs/pull/52/files#diff-52b718f0ee2e4e8580f7155fe871a4267a6427607bf486005bed17325f066a1dR95) with `(None, None)` we actually convert into `ParseEnvVar`. I feel like they are a bit inconsistent. Why is it an error here but not here?

src/unix.rs
BelovDV: This is partially reverted #6, copy of code from old version. Or you mean to insert this in code?

src/unix.rs
weihanglo: >  Is smth bad in calling from_pipe after succeeded from_fifo?

Not really. Just an idea that if we don't need it we don't create it. It also reduces syscalls. Anyway, I am not the maintainer of this project. Just someone cares about it :)

src/unix.rs
weihanglo: I see. Thanks for the reference!

BelovDV:
Thanks for review, I'll wait other's comments.

src/unix.rs
weihanglo: Sounds reasonable! I am sold 😆

Just realized there is no button "Mark as Resolved" for this thread even it was initiated by me 🤔

src/lib.rs
```diff
@@ -151,6 +151,22 @@ struct HelperInner {
...
 
+/// Error type for `from_env` function.
+#[derive(Debug)]
+pub enum ErrFromEnv {
+    /// There isn't env var, that describes jobserver to inherit.
+    IsNotConfigured,
+    /// Cannot connect following this process's environment.
+    PlatformSpecific {
+        /// Error.
+        err: imp::ErrFromEnv,
```
NobodyXu, line 162: Nit: Having os specific exposed doesn't sound great
I wonder if it is a better idea to use `io::Error` here since the return of error here is to improve logging/err msg on failure.

NobodyXu (changes requested):

src/lib.rs
BelovDV: Some errors aren't `io::Error`, wouldn't such error type mislead? I've chose from `imp::ErrFromEnv` and `{ description: &'static str, additional_info: String }`.

The reason of current choice. Os specific errors can be either printed or passed to `match` to be processed. Default `Debug` can be changed. If user want to process `PlatformSpecific`, probably it's better to explicitly state by using `cfg`, that code will deal with platform specific behavior.

If it would be enough in most cases to separate `IsNotConfigured` from other errors, I'd be sure in current variant, but unix `InvalidDescriptor`, according to gnu make, should be considered as a warning, not an error. So if os specific shouldn't be in public interface - I'll change this.

NobodyXu: > Some errors aren't io::Error, wouldn't such error type mislead? I've chose from imp::ErrFromEnv and { description: &'static str, additional_info: String }.

`io::ErrorKind::Others` can be used for these kind of errors and other crates I've seen also use `io::Error` for internal errors, so not exactly misleading.

For example, `tokio::task::JoinError` implements `impl From<JoinError> for std::io::Error`.

[`std::thread::Builder::spawn`](https://doc.rust-lang.org/std/thread/struct.Builder.html#method.spawn) also uses `io::Error` even if creating a new thread isn't an I/O.

> The reason of current choice. Os specific errors can be either printed or passed to match to be processed. Default Debug can be changed. If user want to process PlatformSpecific, probably it's better to explicitly state by using cfg, that code will deal with platform specific behavior.

Does the user really need that much flexibility?
I think if you want to create a new error type for `jobserver`, you'd better create a new `struct Error` that does not expose any public fields for the sake of backwards compatibility and instead only expose methods to return `ErrorKind` and implement `std::error::Error::source` for user to access the underlying source if possible.

> If it would be enough in most cases to separate IsNotConfigured from other errors, I'd be sure in current variant, but unix InvalidDescriptor, according to gnu make, should be considered as a warning, not an error. So if os specific shouldn't be in public interface - I'll change this.

IMO `io::ErrorKind` already provides enough kinds to classify the requirements in #51 

> Error or warn if the jobserver descriptors are open/accessible, but corrupted, i.e. reopened by someone else and no longer refer to pipes 

IMO [`io::ErrorKind::InvalidData`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.InvalidInput) can be used for this.

src/lib.rs
```diff
@@ -151,6 +151,22 @@ struct HelperInner {
...
+        /// Name of gotten env var.
+        env: &'static str,
+        /// Value of gotten env var.
+        var: String,
+    },
+}
+
```
NobodyXu, line 169:
```suggestion
impl std::fmt::Display for ErrFromEnv {
    // ...
}

impl std::error::Error for ErrFromEnv {
    fn source(&self) -> Option<&dyn std::error::Error> {
        match self {
            ErrorFromEnv::PlatformSpecific { err } => Some(err),
            _ => None,
        }
    } 
}
```

Or you can use [`thiserror`](https://docs.rs/thiserror) for convenience, but since jobserver is used in cargo they might not want this.

src/unix.rs
```diff
@@ -25,6 +25,13 @@ pub struct Acquired {
     byte: u8,
 }
 
+#[derive(Debug)]
+pub enum ErrFromEnv {
+    ParseEnvVar,
+    OpenFile(String),
```
NobodyXu, line 31: The filename should be `PathBuf` just in case of non-utf8 characters (jobserver currently uses `env::var` so it is guaranteed to be valid utf8, but IMO it should use `env::var_os` instead for non utf-8 filename).

```suggestion
    OpenFile { filename: PathBuf, err: std::io::Error },
```

src/unix.rs
```diff
@@ -25,6 +25,13 @@ pub struct Acquired {
     byte: u8,
 }
 
+#[derive(Debug)]
```
NobodyXu, line 28: You should also implement `std::fmt::Display` and `std::error::Error`.

src/unix.rs
```diff
@@ -385,8 +397,26 @@ impl Helper {
     }
 }
 
-fn is_valid_fd(fd: c_int) -> bool {
-    unsafe { libc::fcntl(fd, libc::F_GETFD) != -1 }
+fn check_fd(fd: c_int) -> bool {
+    #[cfg(feature = "check_pipe")]
```
NobodyXu, line 401:
```suggestion
    #[cfg(feature = "check_pipe")]
```

IMHO this check should not be optional

src/windows.rs
```diff
@@ -127,17 +133,17 @@ impl Client {
...
         let name = match CString::new(s) {
             Ok(s) => s,
-            Err(_) => return None,
+            Err(_) => return Err(ErrFromEnv::IsNotCString),
         };
 
         let sem = OpenSemaphoreA(SYNCHRONIZE | SEMAPHORE_MODIFY_STATE, FALSE, name.as_ptr());
         if sem.is_null() {
-            None
+            Err(ErrFromEnv::CannotAcquireSemaphore)
```
NobodyXu, line 144: Wouldn't it be better to return `io::Error` here, which will give much better error message?

src/unix.rs
```diff
@@ -25,6 +25,13 @@ pub struct Acquired {
     byte: u8,
 }
 
+#[derive(Debug)]
+pub enum ErrFromEnv {
+    ParseEnvVar,
+    OpenFile(String),
+    InvalidDescriptor(i32, i32),
```
NobodyXu, line 32: It's really confusing what each field is for and what is the reason for the failure.

If you don't want to use `io::Error`, you should at least include a `reason` on why the fd is rejected: Is it non-existent or is it not a pipe?

```suggestion
    InvalidDescriptor { read: std::os::fd::RawFd, write: std::os::fd::RawFd, reason: &'static str },
```

NobodyXu (changes requested):

src/lib.rs
BelovDV: I didn't know enough about `io::Error`. Thanks, I'll change to this.

src/unix.rs
BelovDV: It was removed with one reason. And there isn't such check in gnu make.

I'm not sure about default behavior though.

NobodyXu: [POSIX jobserver](https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html) requires the jobserver on Unix to be implemented with either named pipe or fifo, so if they pass something else, it's incorrect and should be reject, at least that's how I interpret the spec.

Cargo.toml
```diff
@@ -15,7 +15,7 @@ edition = "2018"
 libc = "0.2.50"
 
 [features]
-check_pipe = []
+do_not_check_pipe = []
```
NobodyXu, line 18: This is also not a good idea since features should be additive.

With this do_not_check_pipe feature,if one of the crate enables this, then  every other crate is opt out of checking.

src/unix.rs
```diff
@@ -88,51 +81,41 @@ impl Client {
...
-        let path = match parts.next() {
-            Some(p) => Path::new(p),
-            None => return Some(Err(ErrFromEnv::ParseEnvVar)),
-        };
-        let file = match OpenOptions::new().read(true).write(true).open(path) {
-            Ok(f) => f,
-            Err(e) => return Some(Err(ErrFromEnv::OpenFile(e.to_string()))),
-        };
-        Some(Ok(Client::Fifo {
+        let path = Path::new(parts.next().ok_or(io::Error::new(
```
NobodyXu, line 103: Use ok_or_else to only create io::Error on failure, since io::Error::new does heap allocation internally.

NobodyXu (changes requested):

Cargo.toml
BelovDV: Otherwise if one enables check then one uses smth different from pipe will fail.

alexcrichton:
Given how low-level this crate is I don't think that the Cargo feature added here will be all that useful since if it's in use it's buried so deep and will be difficult to configure.

Can you explain a bit more what the error handling around pipe detection is doing? Why is `fcntl` not sufficient and what is `fstat` catching otherwise? Is that catching real issues happening in the wild and allowing callers to recover?

petrochenkov:
>I don't think that the Cargo feature added here will be all that useful

+1

> Can you explain a bit more what the error handling around pipe detection is doing? Why is `fcntl` not sufficient and what is `fstat` catching otherwise?

The pipe check is the same thing that jobserver-rs did before https://github.com/alexcrichton/jobserver-rs/pull/6.

> Is that catching real issues happening in the wild and allowing callers to recover?

It catches a real issue that happens in the wild.
The caller won't be able to recover, but it will be able to report a useful error that will allow the "second order caller" (build engineer) to fix the build environment.

Cargo.toml
petrochenkov: If checking or not checking the pipe-ness should be configurable, then it should probably be a runtime option passed to `from_env`.

Not checking is the GNU Make behavior, checking catches errors in practice.

src/lib.rs
```diff
@@ -227,28 +269,24 @@ impl Client {
...
-            Some(i) => i,
-            None => {
-                arg = "--jobserver-auth=";
-                match var.find(arg) {
-                    Some(i) => i,
-                    None => return None,
-                }
-            }
-        };
+    pub unsafe fn from_env() -> Result<Client, ErrFromEnv> {
```
petrochenkov, line 272: I'd suggest to keep the simple version of the function alone, and implement it in terms of the "extended" function.
It will also avoid breaking changes to the crate.
Not everyone needs the new diagnostics immediately.

alexcrichton:
If this is reverting back, is the use case in https://github.com/alexcrichton/jobserver-rs/pull/6 no longer valid?

For "catching things in the wild" I meant moreso the new logic to detect a valid file descriptor, less so "should this return errors at all" because that's naturally a "sure this should happen when the effort is put in".

src/lib.rs
```diff
@@ -151,6 +151,44 @@ struct HelperInner {
...
 
+/// Error type for `from_env` function.
+#[derive(Debug)]
+pub enum ErrFromEnv {
+    /// There isn't env var, that describes jobserver to inherit.
+    IsNotConfigured,
+    /// Cannot connect following this process's environment.
+    PlatformSpecific {
+        /// Error.
+        err: io::Error,
```
petrochenkov, line 162: @weihanglo 
Could you try this interface in cargo before it's in a "stable" release of jobserver-rs?

It's possible that in practice we'll need to discern between different kinds of erros (e.g. "descriptors are closed" and everything else) to avoid bothering users with warnings too often.

This PR, however, merges all the errors into a single `io::Error` with multiple vague  `ErrorKind`s.
I'm not sure users will be able to use it reliably without relying on "string typing" (inspecting string descriptions returned by errors) or implementation details ("we know that closed descriptors are `ErrorKind::InvalidData` by reading source code of jobserver-rs").
(I don't like this part and I wish I read this PR in detail before its merge.)

weihanglo: Sure. I'll try taking a look this week.

NobodyXu: > implementation details ("we know that closed descriptors are ErrorKind::InvalidData by reading source code of jobserver-rs").

Maybe jobserver can have the error kind documented on docs.rs?

weihanglo: Sorry for the late response. I think the change works fine with Cargo. Cargo can just [ignore the jobserver if there are bad pipes](https://github.com/rust-lang/cargo/blob/7bf43f028ba5eb1f4d70d271c2546c38512c9875/src/cargo/util/config/mod.rs#L258-L262). No harm to get this patch released on Cargo side I believe :)

PR #54, opened 2023-04-11
Error type for `from_env_ext` function

belovdv:
Continue implementing #51. The previous PR wasn't finished and cleaned up.

If we want to discern different errors and treat them differently, more specific return type would be more convenient.

If this's going to be merged, probably there'll be some changes from current version. If it's possible, i'd like to clean up results before merging.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,78 @@
+#[cfg(unix)]
+type RawFd = std::os::fd::RawFd;
+#[cfg(not(unix))]
+type RawFd = std::convert::Infallible;
+
+/// Error type for `from_env_ext` function.
+#[derive(Debug)]
+pub struct FromEnvError {
+    pub(crate) inner: FromEnvErrorInner,
+}
+
+/// Kind of an error returned from `from_env_ext` function.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum FromEnvErrorKind {
+    /// There is no environment variable that describes jobserver to inherit.
+    NoEnvVar,
+    /// Cannot parse jobserver environment variable value, incorrect format.
+    CannotParse,
+    /// Cannot open path or name from the jobserver environment variable value.
+    CannotOpenPath,
+    /// Cannot open file descriptor from the jobserver environment variable value.
+    CannotOpenFd,
+    /// File descriptor from the jobserver environment variable value is not a pipe.
+    NotAPipe,
+    /// Jobserver inheritance is not supported on this platform.
+    Unsupported,
+}
+
+impl FromEnvError {
+    /// Get the error kind.
+    pub fn kind(&self) -> FromEnvErrorKind {
+        match self.inner {
+            FromEnvErrorInner::NoEnvVar => FromEnvErrorKind::NoEnvVar,
+            FromEnvErrorInner::CannotParse(_) => FromEnvErrorKind::CannotParse,
+            FromEnvErrorInner::CannotOpenPath(..) => FromEnvErrorKind::CannotOpenPath,
+            FromEnvErrorInner::CannotOpenFd(..) => FromEnvErrorKind::CannotOpenFd,
+            FromEnvErrorInner::NotAPipe(..) => FromEnvErrorKind::NotAPipe,
+            FromEnvErrorInner::Unsupported => FromEnvErrorKind::Unsupported,
+        }
+    }
+}
+
+impl std::fmt::Display for FromEnvError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match &self.inner {
+            FromEnvErrorInner::NoEnvVar => write!(f, "there is no environment variable that describes jobserver to inherit"),
+            FromEnvErrorInner::CannotParse(s) => write!(f, "cannot parse jobserver environment variable value: {s}"),
+            FromEnvErrorInner::CannotOpenPath(s, err) => write!(f, "cannot open path or name {s} from the jobserver environment variable value: {err}"),
+            FromEnvErrorInner::CannotOpenFd(fd, err) => write!(f, "cannot open file descriptor {fd} from the jobserver environment variable value: {err}"),
+            FromEnvErrorInner::NotAPipe(fd, None) => write!(f, "file descriptor {fd} from the jobserver environment variable value is not a pipe"),
+            FromEnvErrorInner::NotAPipe(fd, Some(err)) => write!(f, "file descriptor {fd} from the jobserver environment variable value is not a pipe: {err}"),
+            FromEnvErrorInner::Unsupported => write!(f, "jobserver inheritance is not supported on this platform"),
+        }
+    }
+}
+impl std::error::Error for FromEnvError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match &self.inner {
+            FromEnvErrorInner::CannotOpenPath(_, err) => Some(err),
+            FromEnvErrorInner::NotAPipe(_, Some(err)) | FromEnvErrorInner::CannotOpenFd(_, err) => {
+                Some(err)
+            }
+            _ => None,
+        }
+    }
+}
+
+#[allow(dead_code)]
+#[derive(Debug)]
+pub(crate) enum FromEnvErrorInner {
+    NoEnvVar,
+    CannotParse(String),
+    CannotOpenPath(String, std::io::Error),
+    CannotOpenFd(RawFd, std::io::Error),
+    NotAPipe(RawFd, Option<std::io::Error>),
+    Unsupported,
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,10 +82,12 @@
 #![doc(html_root_url = "https://docs.rs/jobserver/0.1")]
 
 use std::env;
+use std::ffi::OsString;
 use std::io;
 use std::process::Command;
 use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 
+mod error;
 #[cfg(unix)]
 #[path = "unix.rs"]
 mod imp;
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -151,40 +152,80 @@ struct HelperInner {
...
-    },
+pub struct FromEnv {
+    /// Result of trying to get jobserver client from env.
+    pub client: Result<Client, FromEnvError>,
+    /// Name of the environment variable.
+    /// Empty if no relevant environment variable is found.
+    pub var_name: &'static str,
+    /// Value of the environment variable.
+    /// Empty if no relevant environment variable is found.
+    pub var_value: OsString,
```
NobodyXu, line 165: Why not use `Option` to represent no env variable is found and group them together as one field?
```suggestion
    /// Name and value of the environment variable.
    /// `None` if no relevant environment variable is found.
    pub var: Option<(&'static str, OsString)>,
```

src/lib.rs
```diff
@@ -151,40 +152,80 @@ struct HelperInner {
...
+#[non_exhaustive]
+pub enum FromEnvError {
+    /// There is no environment variable that describes jobserver to inherit.
+    NoEnvVar,
+    /// Cannot parse jobserver environment variable value, incorrect format.
+    CannotParse(String),
+    /// Cannot open path or name from the jobserver environment variable value.
+    CannotOpenPath(String, io::Error),
+    /// Cannot open file descriptor from the jobserver environment variable value.
+    CannotOpenFd(std::ffi::c_int, io::Error),
```
NobodyXu, line 179: Is it really ok to expose unix-specific error variant?

```suggestion
    #[cfg(unix)]
    CannotOpenFd(std::os::fd::RawFd, io::Error),
```

src/lib.rs
```diff
@@ -151,40 +152,80 @@ struct HelperInner {
...
+/// Error type for `from_env_ext` function.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum FromEnvError {
+    /// There is no environment variable that describes jobserver to inherit.
+    NoEnvVar,
+    /// Cannot parse jobserver environment variable value, incorrect format.
+    CannotParse(String),
+    /// Cannot open path or name from the jobserver environment variable value.
+    CannotOpenPath(String, io::Error),
```
NobodyXu, line 177:
```suggestion
    CannotOpenPath(PathBuf, io::Error),
```

NobodyXu (changes requested):

src/lib.rs
petrochenkov: Code using jobserver-rs will be target-independent, most likely, so it's probably simpler to avoid cfgs.

src/lib.rs
petrochenkov: `FromEnvError::NoEnvVar` also indicates that there's no env var and these fields can be ignored.
If there's also an `Option` they can in theory go into conflict, and you'll have to unwrap, so I guess no `Option` is simpler.

("Empty if no relevant environment ..." could easily be "Unspecified if no relevant environment ...")

src/lib.rs
NobodyXu: Sounds reasonable , I thought about using OsString for a moment but then you would still need to convert to string in formatting.

src/lib.rs
NobodyXu: Maybe it's better to use Result<(Client, EnvVariable), FromEnvError> and also includes EnvVariable in FromEnvError except for variant NoEmvVar and NotSupporte?

    struct EnvVariable {
        name: &'static str,
        value: OsString,
    }

src/lib.rs
NobodyXu: Even so, I think it's better to use std::os::fd::RawFd instead of c_int

NobodyXu: Thanks, it's a shame that we have to use a raw integer type here and expose what I considered to be implementation detail, but I can't think of something better right now.

NobodyXu: @petrochenkov Could we merge CannotOpen{Path, Fd} into one and make the descriptor (path/fd) opaque?

I don't think cargo would go to the extent of inspecting which path/fd that open failed, so maybe we don't need to expose such impl detail and os specific detail?

src/lib.rs
petrochenkov: Actually maybe it's reasonable to make it `OsString` because the env var values are `OsString`.

It's hard to parse `OsString`s without converting to `String` right now because methods like `contains` or `starts_with` are lacking, but it would be more future compatible.

src/lib.rs
NobodyXu: > But it could probably be solved by adding a new `FromEnvError` variant too.

IMO adding a new variant is better than exposing implementation detail and let the user check it themselves.

src/lib.rs
belovdv: In case of negative file descriptor `from_env_ext` may want to return `Client` without environment variables. Maybe it's better to make result and env independent like `(Result<Client, FromEnvError>, Option<EnvVariable>)`?

src/lib.rs
belovdv: If fd is going to be used only for error message, would it be enough to note is it write or read descriptor, without actual number that is in environment variable value? imho negative file descriptors and things like that should be handled within this crate.

src/lib.rs
NobodyXu: > In case of negative file descriptor `from_env_ext` may want to return `Client` without environment variables.

I don't `from_env_ext` should return Client on negative file descriptor.

Negative file descriptor means the jobserver is disabled.

In that case, the user could create a client using Client::nee

src/lib.rs
petrochenkov: If people are not against multiplying public types, then we can go with
```rust
pub enum FromEnvError {
    NoEnvVar,
    CannotParse(CannotParseError),
    CannotOpenPath(CannotOpenPathError),
    CannotOpenFd(CannotOpenFdError),
    NotAPipe(NotAPipeError),
    Unsupported,
}
```
where `{CannotParse,CannotOpenPath,CannotOpenFd,NotAPipe}Error` are all opaque structures.

This would be future-compatible with keeping all kind of stuff inside them.

petrochenkov: >where `{CannotParse,CannotOpenPath,CannotOpenFd,NotAPipe}Error` are all opaque structures

They can even be made unnameable for a start.
```
mod error  {
    pub struct CannotParseError { ... }
}

pub enum FromEnvError {
    CannotParse(CannotParseError),
    ...
}
```

NobodyXu: If that's the case, why not make `FromEnvError` and opaque type and only expose:

```rust
pub enum FromEnvErrorKind {
    CannotParse,
    CannotOpenPath,
    CannotOpenFd,
    NotAPipe,
}

impl FromEnvError {
    pub fn kind(&self) -> FromEnvErrorKind;
}
```

petrochenkov: Yeah, that should be fine too.

src/error.rs
```diff
@@ -0,0 +1,87 @@
+/// Opaque error type for `from_env_ext` function.
+#[derive(Debug)]
+pub struct FromEnvError {
+    pub(crate) inner: FromEnvErrorInner,
+}
+
+/// Public simplified representation of error type for `from_env_ext` function.
```
petrochenkov, line 7:
```suggestion
/// Kind of an error returned from `from_env_ext` function.
```
You are writing about implementation details in documentation for users.

src/error.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Cannot open file descriptor from the jobserver environment variable value.
+    CannotOpenFd,
+    /// File descriptor from the jobserver environment variable value is not a pipe.
+    NotAPipe,
+    /// Jobserver inheritance is not supported on this platform.
+    Unsupported,
+}
+
+impl FromEnvError {
+    /// Get public version of error type.
```
petrochenkov, line 26:
```suggestion
    /// Get the error kind.
```

src/error.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum FromEnvErrorInner {
+    NoEnvVar,
+    CannotParse(String),
+    #[cfg(any(unix, windows))]
+    CannotOpenPath(String, std::io::Error),
```
petrochenkov, line 80: `#[cfg(unix)]` on `CannotOpenFd`/`NotAPipe` allow using `RawFd`, but cfgs on this variant and `Unsupported` do not bring any benefit.

src/lib.rs
```diff
@@ -234,10 +226,10 @@ impl Client {
     ///
     /// # Return value
     ///
+    /// `FromEnv` contains result and relevant `var_name` with `var_value`.
```
petrochenkov, line 229: >`var_name` with `var_value`

Outdated comment

src/error.rs
```diff
@@ -0,0 +1,79 @@
...
+            FromEnvErrorInner::NotAPipe(_, Some(err)) | FromEnvErrorInner::CannotOpenFd(_, err) => {
+                Some(err)
+            }
+            _ => None,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum FromEnvErrorInner {
```
petrochenkov, line 72:
```suggestion
enum FromEnvErrorInner {
```

src/error.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[cfg(unix)]
+pub type RawFd = std::os::fd::RawFd;
+#[cfg(not(unix))]
+pub type RawFd = std::convert::Infallible;
+
+/// Error type for `from_env_ext` function.
+#[derive(Debug)]
+pub struct FromEnvError {
+    pub(crate) inner: FromEnvErrorInner,
```
petrochenkov, line 11:
```suggestion
    inner: FromEnvErrorInner,
```

src/error.rs
```diff
@@ -0,0 +1,79 @@
+#![allow(dead_code)]
+
+#[cfg(unix)]
+pub type RawFd = std::os::fd::RawFd;
```
petrochenkov, line 4:
```suggestion
type RawFd = std::os::fd::RawFd;
```

src/error.rs
```diff
@@ -0,0 +1,79 @@
...
+#[cfg(not(unix))]
+pub type RawFd = std::convert::Infallible;
```
petrochenkov, line 6:
```suggestion
enum RawFd {} // uninhabited
```

src/error.rs
```diff
@@ -0,0 +1,79 @@
+#![allow(dead_code)]
```
petrochenkov, line 1: This should no longer be necessary?
(If it's still necessary it should be put on something more specific than a module.)

petrochenkov (approved):

petrochenkov:
ping @alexcrichton for merging
ping @weihanglo in case you have any additional comments

(So far this PR has consensus between me, @NobodyXu and @belovdv.)

alexcrichton (commented):
Sorry I haven't been following the discussion here and it seems like there's a good deal, but this to me feels like it's dialing the wrong knob per se up to 11. The `FromEnv` structure feels pretty unidiomatic as it's a struct-containing-a-`Result` which breaks `?` usage  and requires filling out fields that may not always make the most sense. I can sort of see how it's conveying more information but I would not personally settle on this particular design.

On one hand this is fine. This isn't exactly a heavily used crate by everyone in every project so it's not like something that's a bit unidiomatic will make that much of a difference. On the other hand, though, this is not code that I would want to maintain. I would prefer to improve the idioms here and figure out a different way which I'm more comfortable with which achieves the same goal. Unfortunately I don't have a ton of time or energy to go through this process. At this point it may be best for me to hand off this crate to someone else, perhaps `rust-lang`, since it's depended on by `rustc` and `cargo`. I at the same time don't want to block progress on feature work here.

petrochenkov:
>The FromEnv structure feels pretty unidiomatic as it's a struct-containing-a-Result which breaks ? usage and requires filling out fields that may not always make the most sense.

`?` is not exactly broken because `Client::from_env_ext().client?` would still work.

But it's not too important anyway, this function is called ~once in a program, and if someone is calling `Client::from_env_ext()` instead of `Client::from_env()` he's likely going to inspect and/or log all those errors and env vars.

`FromEnv::var` is not inside the `Result` because it may exist regardless of the client being successfully constructed or not.

alexcrichton:
I'll reiterate the second part of my comment above. I realize that this is a minor crate which is not really all that heavily used. That being said I do not want to personally maintain the idioms introduced here, nor do I necessarily have the time or energy to walk through the review process towards something I would be willing to maintain. If updating this PR requires lots of feedback from me, or if this PR would rather not be updated, then I think it's best to look for a different maintainer of this crate to transfer ownership to.

petrochenkov:
This PR would rather not be updated.
I'll make a proposal to move this crate to rust-lang organization.

src/lib.rs
```diff
@@ -269,27 +261,42 @@ impl Client {
...
+            .find_map(|p| p)
+        {
+            Some((env, var_os)) => (env, var_os),
+            None => return FromEnv::new_err(FromEnvErrorInner::NoEnvVar, "", Default::default()),
+        };
+
+        let var = match var_os.to_str() {
+            Some(var) => var,
+            None => {
+                let err = FromEnvErrorInner::CannotParse("not a valid UTF-8".to_string());
```
thomcc, line 277:
```suggestion
                let err = FromEnvErrorInner::CannotParse("not valid UTF-8".to_string());
```

thomcc (approved):
Looks fine with typo fixed.

PR #56, opened 2023-07-26
Error type for from_env_ext function: update

belovdv:
Continue #51 implementation

It's [possible](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html) that env var exists but doesn't contain jobserver. We'd like to discern this situation from `CannotParse` error.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -15,6 +15,9 @@ pub struct FromEnvError {
 pub enum FromEnvErrorKind {
     /// There is no environment variable that describes jobserver to inherit.
     NoEnvVar,
+    /// There is no jobserver in the environment variable.
+    /// Variables associated with Make can be used for passing data other than jobserver info.
+    NoJobserver,
     /// Cannot parse jobserver environment variable value, incorrect format.
     CannotParse,
     /// Cannot open path or name from the jobserver environment variable value.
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -15,6 +15,8 @@ pub struct FromEnvError {
 pub enum FromEnvErrorKind {
     /// There is no environment variable that describes jobserver to inherit.
     NoEnvVar,
+    /// There is no jobserver in environment variable.
```
petrochenkov, line 18:
```suggestion
    /// There is no jobserver in the environment variable.
    /// Variables associated with Make can be used for passing data other than jobserver info.
```

src/error.rs
```diff
@@ -45,6 +48,7 @@ impl std::fmt::Display for FromEnvError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self.inner {
             FromEnvErrorInner::NoEnvVar => write!(f, "there is no environment variable that describes jobserver to inherit"),
+            FromEnvErrorInner::NoJobserver => write!(f, "there is no jobserver in environment variable"),
```
petrochenkov, line 51:
```suggestion
            FromEnvErrorInner::NoJobserver => write!(f, "there is no `--jobserver-fds=` or `--jobserver-auth=` in the environment variable"),
```

src/error.rs
petrochenkov: The second sentence wasn't added.

PR #57, opened 2023-08-03
feat: Make `Client::from_env` safe to call for any number of times

NobodyXu:
So that it can be used in different crates without causing UB.

## Motivation

[`opencv-rust`](https://github.com/twistedfall/opencv-rust) uses `jobserver::Client::from_env` in its `build.rs` to parallelize generation of bindings, then call `cc::Build::compile` with features `cc/parallel` enabled it will also call `jobserver::Client::from_env` and creates a UB https://github.com/rust-lang/cc-rs/issues/844#issuecomment-1663960861

Workaround for this will require either wrapping `jobserver::Client` with `std::mem::ManuallyDrop`, or switching to my fork [`jobslot`](https://docs.rs/jobslot) which supports calling `Client::from_env` unlimited number of times.

This PR thus port the change to upstream so that any crate using `jobserver` can benefit from this and can stop worrying about UB caused by calling `Clieng::from_env` multiple times in different crates.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -231,13 +231,6 @@ impl Client {
     /// result with the connected client will be returned. In other cases
     /// result will contain `Err(FromEnvErr)`.
     ///
-    /// Note that on Unix the `Client` returned **takes ownership of the file
-    /// descriptors specified in the environment**. Jobservers on Unix are
-    /// implemented with `pipe` file descriptors, and they're inherited from
-    /// parent processes. This `Client` returned takes ownership of the file
-    /// descriptors for this process and will close the file descriptors after
-    /// this value is dropped.
-    ///
     /// Additionally on Unix this function will configure the file descriptors
     /// with `CLOEXEC` so they're not automatically inherited by spawned
     /// children.
@@ -256,11 +249,7 @@ impl Client {
     /// make sure to take ownership properly of the file descriptors passed
     /// down, if any.
     ///
-    /// It's generally unsafe to call this function twice in a program if the
-    /// previous invocation returned `Some`.
-    ///
-    /// Note, though, that on Windows it should be safe to call this function
-    /// any number of times.
+    /// It is ok to call this function any number of times.
     pub unsafe fn from_env_ext(check_pipe: bool) -> FromEnv {
         let (env, var_os) = match ["CARGO_MAKEFLAGS", "MAKEFLAGS", "MFLAGS"]
             .iter()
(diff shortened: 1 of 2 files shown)


src/unix.rs
```diff
@@ -435,6 +436,14 @@ unsafe fn fd_check(fd: c_int, check_pipe: bool) -> Result<(), FromEnvErrorInner>
     }
 }
 
+fn clone_fd_and_set_cloexec(fd: c_int) -> Result<File, FromEnvErrorInner> {
```
carbotaniuman, line 439: You can use `BorrowedFd::try_clone_to_owned` and then `File`'s `From<OwnedFd>` here. Not sure if it's shorter but probably more clear safety-wise.

src/lib.rs
```diff
@@ -256,11 +256,7 @@ impl Client {
     /// make sure to take ownership properly of the file descriptors passed
```
carbotaniuman, line 256: I can't target in on GitHub, but the safety comment above this still states that it takes ownership, which is now incorrect.

src/unix.rs
NobodyXu: Thanks, but this PR is already merged.

I suggest opening a new PR for this.

NobodyXu: Sorry for the confusing comment (which I have deleted), I will apply your suggestion once I'm free.

NobodyXu: @carbotaniuman BorrowedFd is introdiced in 1.63, but cc (which uses this crate) has msrv set to 1.46 https://github.com/rust-lang/cc-rs/blob/360754b91441395713c24db6e5a5e3a91b298527/.github/workflows/main.yml#L124

Unfortunately I can't apply your suggestion due to this.

src/lib.rs
NobodyXu: Thank you!
I've fixed the doc.

src/unix.rs
weihanglo, line 439: Does this really set `FD_CLOEXEC`?

weihanglo: Looks like jobserver has already used some new API and was bumped to 1.63 a while back, so I guess we could use that?

NobodyXu: Yes, it `File::try_clone` calls [`BorrowedFd<'_>::try_clone_to_owned`](https://doc.rust-lang.org/src/std/os/fd/owned.rs.html#96-113) which sets close-on-exec.

NobodyXu: Thanks I didn't notice that, will apply your suggestion.

NobodyXu: Updated.

src/unix.rs
```diff
@@ -444,6 +445,14 @@ unsafe fn fd_check(fd: c_int, check_pipe: bool) -> Result<(), FromEnvErrorInner>
...
+    // Safety: fd is a valid fd dand it remains open until returns
```
weihanglo, line 449:
```suggestion
    // Safety: fd is a valid fd and it remains open until returns
```

weihanglo (approved):
Looks good to me. Thanks!

src/lib.rs
```diff
@@ -293,6 +282,19 @@ impl Client {
...
+    /// # Safety
+    ///
+    /// This function is `unsafe` to call on Unix specifically as it
+    /// transitively requires usage of the `from_raw_fd` function, which is
+    /// itself unsafe in some circumstances.
+    ///
+    /// It's recommended to call this function very early in the lifetime of a
+    /// program before any other file descriptors are opened. That way you can
+    /// make sure to take ownership properly of the file descriptors passed
+    /// down, if any.
```
weihanglo, line 295: nit: How about this?

```suggestion
    /// # Safety
    ///
    /// See the "Safety" section in [`Client::from_env_ext`].
```

weihanglo:
I don't think it worth more back-and-forth. Let's merge this and improve comments later!

PR #59, opened 2023-10-07
chore: set MSRV to 1.66 and edition to 2021

weihanglo:
Some new API were already used

* format arg capture (1.58)
* `std::os::fd` (1.66)

A bump is inevitably coming.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,8 @@ documentation = "https://docs.rs/jobserver"
 description = """
 An implementation of the GNU make jobserver for Rust
 """
-edition = "2018"
+edition = "2021"
+rust-version = "1.66"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.50"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,23 +7,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -9,7 +9,8 @@ documentation = "https://docs.rs/jobserver"
 description = """
 An implementation of the GNU make jobserver for Rust
 """
-edition = "2018"
+edition = "2021"
+rust-version = "1.66.1"
```
epage, line 13: nit: I leave the patch version unspecified because it causes unnecessary churn to MSRVs when a dependency bumps the patch

weihanglo: Thanks. Update.

weihanglo:
@petrochenkov, do you know who's the owner of this repo, and how the release process look like?

petrochenkov:
@weihanglo 
I can publish new versions of this crate now.

petrochenkov:
Could you add a CI check for the minimum supported version?
Example: https://github.com/mozilla/cbindgen/blob/master/.github/workflows/cbindgen.yml#L32-L38

weihanglo:
Updated. Let me know if it is too excessive, and we can slim it down a bit.

petrochenkov:
Tests are hanging since yesterday, so something is probably wrong with CI changes (missing `-latest` on macos and windows?).

PR #62, opened 2023-11-22
Reduce MSRV to 1.63 with minor change in RawFd type

hoxxep:
I noticed that a patch version increment has caused other crates MSRV values to change, such as the zstd crate: https://github.com/gyscos/zstd-rs/issues/253

This MR changes the RawFd type to one present in older versions of rust, so that the MSRV can be reduced to 1.63.

My understanding is that this type should be equivalent because it's gated against `#[cfg(unix)]`, and so the benefits of `std::os::fd::RawFd` also supporting WASM is negated. https://github.com/rust-lang/rust/issues/98699

In theory the MSRV of the crate library is 1.60, but there are dev-dependencies that require 1.63 which would break the MSRV testing workflow (rustix, required by tempfile). 

If we insist on keeping the usage of `std::os::fd::RawFd`, I can feed back to the zstd crate and ask them to bump their MSRV instead. Thank you!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        rust: ["1.60", stable, beta, nightly]
+        rust: ["1.63", stable, beta, nightly]
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
     - uses: actions/checkout@master
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        rust: ["1.66", stable, beta, nightly]
+        rust: ["1.60", stable, beta, nightly]
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
     - uses: actions/checkout@master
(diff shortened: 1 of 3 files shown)


epage:
Not speaking to the value of this change but to the motivation:

> I noticed that a patch version increment has caused other crates MSRV values to change, such as the zstd crate: https://github.com/gyscos/zstd-rs/issues/253


A change in MSRV in this crate does not affect the MSRV of any other crate unless they take an action like bump their version requirement. You can maintain a valid dependency tree for your MSRV via `Cargo.lock` / `cargo update`. You can even use `cargo +nightly generate-lockfile -Zmsrv-policy`. We are discussing stabilization on internals.

epage:
MSRV bumps are generally recognized as non-breaking.

From a tooling perspective, minor and patch are equivalent.

Also, a 1 year old MSRV covers 99% of requests to crates.io.

hoxxep:
@epage I've copied our discussion into your internals thread since it seemed a better place to discuss than this PR 😊 https://internals.rust-lang.org/t/pre-rfc-msrv-aware-resolver/19871/50

Agreed the current interpretation is that MSRV bumps are considered non-breaking. They clearly are breaking for some users though, and so perhaps with the MSRV aware RFC we could change the precedent to consider them "semi-breaking" changes. An MSRV bump doesn't require a major version bump, but crates could benefit by leaving room in the versioning scheme to have the option to release patches on the old MSRV for security fixes.

Regarding the crates.io request metrics, I don't doubt the rust community is more up to date than C++ users, but I'd note that legacy/enterprise users will often have their own crates.io mirrors and forked repositories. So crates.io metrics could be skewed in favour of smaller non-legacy projects.

PR #74, opened 2024-03-02
CI: Improve msrv checking and speedup CI

NobodyXu:
Improve msrv CI:
 - Run `cargo +nightly -Zminimal-versions update` before checking for msrv, to create a lockfile containing the minimal versions of dependencies
 - Run `cargo-check` to check for msrv (in a new job), running `cargo-test` would also pull in test dependencies and test code

This PR also speeds up the CI from 5m down to 1m (when make and dependencies are cached).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,6 +18,8 @@ jobs:
       run: rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
       shell: bash
 
+    - run: cargo +stable update
+
     - uses: Swatinem/rust-cache@v2
 
     - run: cargo test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         rust: ["1.63", stable, beta, nightly]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-14, windows-latest]
     steps:
     - uses: actions/checkout@master
     - name: Install Rust (rustup)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@master
     - name: Install Rust (rustup)
       run: |
-        rustup update ${{ matrix.rust }} --no-self-update
+        rustup toolchain install ${{ matrix.rust }} --no-self-update --profile minimal
         rustup default ${{ matrix.rust }}
       shell: bash
 
@@ -93,8 +93,9 @@ jobs:
     - uses: actions/checkout@master
     - name: Install Rust (rustup)
       run: |
-        rustup update $MSRV --no-self-update
+        rustup toolchain install $MSRV --no-self-update --profile minimal
         rustup default $MSRV
+        rustup toolchain install nightly --no-self-update --profile minimal
       shell: bash
 
     - run: cargo +nightly update -Zminimal
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,25 +17,34 @@ jobs:
 
     - run: cargo test
 
+    - name: Cache make compiled
+      if: ${{ !startsWith(matrix.os, 'windows') }}
+      id: cache-make
+      uses: actions/cache@v4
+      with:
+        path: /usr/local/bin/make
+        key: ${{ runner.os }}-make-4.4.1
+
     # Compile it from source (temporarily)
     - name: Make GNU Make from source
-      if: ${{ !startsWith(matrix.os, 'windows') }}
+      if: ${{ !startsWith(matrix.os, 'windows') }} && steps.cache-make.outputs.cache-hit != 'true'
       env:
         VERSION: "4.4.1"
       shell: bash
       run: |
-        wget -q "https://ftp.gnu.org/gnu/make/make-${VERSION}.tar.gz"
-        tar zxf "make-${VERSION}.tar.gz"
+        curl "https://ftp.gnu.org/gnu/make/make-${VERSION}.tar.gz" | tar xz
         pushd "make-${VERSION}"
         ./configure
-        make
+        make -j 4
         popd
-        cp -rp "make-${VERSION}/make" .
+        cp -p "make-${VERSION}/make" /usr/local/bin
+
     - name: Test against GNU Make from source
       if: ${{ !startsWith(matrix.os, 'windows') }}
       shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -63,3 +81,25 @@ jobs:
...
           git push -f -q https://git:${{ secrets.github_token }}@github.com/${{ github.repository }} HEAD:gh-pages
         if: github.event_name == 'push' && github.event.ref == 'refs/heads/master'
+
+  msrv:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-14, windows-latest]
+    env:
+      MSRV: 1.63
```
epage, line 91: If you use `cargo-hack`, you don't need to track the MSRV in yet-another-place

See
- https://github.com/rust-lang/cargo/blob/b4bb1eacdb73f9ca3eb9d878e848e69b64db656d/.github/workflows/main.yml#L269-L270
- https://github.com/rust-lang/cargo/pull/13523

.github/workflows/main.yml
```diff
@@ -63,3 +81,25 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@master
+    - name: Install Rust (rustup)
+      run: |
+        rustup toolchain install $MSRV --no-self-update --profile minimal
+        rustup default $MSRV
+        rustup toolchain install nightly --no-self-update --profile minimal
+      shell: bash
+
+    - run: cargo +nightly update -Zminimal-versions
```
epage, line 101: Please instead commit your lockfile

NobodyXu: I'm not sure committing lockfile is better than `-Zminimal-versions`

`-Zminimal-versions` make sure the minimal version of dependency is used, committing a lockfile will then require user to maintain it whenever they update their dependencies and re-run `cargo +nightly -Zminimal-versions update` locally.

.github/workflows/main.yml
NobodyXu: Thanks, I've updated the PR.

.github/workflows/main.yml
riking: The convention here is to check in a `Cargo.minimal-versions.lock` file or `Cargo.msrv.lock` and use that in this CI job instead

NobodyXu: Thanks, I've updated the PR to use `Cargo.msrv.lock`

epage: No, please do not do that either!  That significantly increases maintenance / degrades the contributor's experience.

For why to commit your lockfile, see https://doc.rust-lang.org/stable/cargo/faq.html#why-have-cargolock-in-version-control

Maintaining a `Cargo.msrv.lock` means you need to have a `--locked` job for that and anytime you use it (or edit `Cargo.toml`), you need to juggle it in and out of your regular `Cargo.lock`.

In the MSRV RFC, I split this up into [two extended MSRV workflows](https://rust-lang.github.io/rfcs/3537-msrv-resolver.html#extended-msrv) and found very little motivation to justify the two-lockfile solution.  Just keep your `Cargo.lock` as MSRV compatible and have a non-blocking "latest versions" verification job.

NobodyXu: I've renamed `Cargo.msrv.lock` back to `Cargo.lock` and commit it.

Please review again if you have time @epage

.github/workflows/main.yml
```diff
@@ -63,3 +81,19 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@master
+    - name: Install Rust (rustup)
+      run: rustup toolchain install nightly --no-self-update --profile minimal
+      shell: bash
+
+    - uses: taiki-e/install-action@cargo-hack
+    - uses: Swatinem/rust-cache@v2
+
+    - run: cargo hack check --lib --rust-version --ignore-private
```
epage, line 99: This needs a `-locked` flag

epage, line 99: If the regular test job is not testing with the old dependencies, this job should

NobodyXu: I can add that back, but that would mean our `dev-dependencies` would also need to be below the msrv

epage: Alternatively, your primary test jobs could test with your lockfile and then this just needs to verify things compile.

(your primary test jobs should be testing with your lockfile anyways because you should isolate external factors from failing a PR build)

NobodyXu: Updated tests and msrv to use `--locked`

NobodyXu: Updated tests and msrv to use `--locked`

NobodyXu: Also fixed the testing by removing `dev-dependencies` that are too old and upgrade libc to 0.2.72

epage (approved):
For CI.  I don't deal enough with this to speak to the test changes

Cargo.lock
```diff
@@ -0,0 +1,99 @@
...
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dfc5b3ce5d5ea144bb04ebd093a9e14e9765bcfec866aecda9b6dec43b3d1e24"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
+name = "tempfile"
+version = "3.0.2"
```
weihanglo, line 68: [tempfile@3.10.1](https://crates.io/crates/tempfile/3.10.1) looks fine with our MSRV (1.63.0) and can remove some other old dependencies like `fuchsia-zircon`.

weihanglo (commented):
Thanks for working on this. Glad to see the huge improvement on CI time!

I'd love to have a clearer git history. It doesn't mean we need to squash commits. Just let it make more sense and remove some intermediate commits during the development.

tests/client.rs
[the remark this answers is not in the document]
NobodyXu: Unfortunately not possible.

It used to use deprecated version of tokio-core and futures, when creating a minimal lockfile out of it, it pullsin libc 0.1 which causes compilation failure.

I have to rip them out and use thread spawning instead.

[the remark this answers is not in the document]
weihanglo: Understood. Thank you!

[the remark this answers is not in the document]
weihanglo: BTW, my original proposal is we can land a PR switching to threads first then this PR. That's not really a blocker though.

Cargo.lock
NobodyXu: Thanks, updated

weihanglo (approved):
Given both CI and test are reviewed, I am going to merge this. Thank you!

PR #77, opened 2024-04-11
Fix warnings from rustc and clippy

petrochenkov:


src/unix.rs
```diff
@@ -291,7 +291,7 @@ impl Client {
         {
             let read = self.read().as_raw_fd();
             loop {
-                match non_blocking_read(read, &mut buf) {
+                match non_blocking_read(read, &buf) {
```
NobodyXu, line 294: This was an error in programming: `non_blocking_read` definitely needs mutable reference to buffer.

cc @weihanglo Can you fix this in your PR (opening another PR would just create merge conflicts with yours)

weihanglo: Feel free to open a PR, since they fix different issues. Fine with resolving PR conflicts for that reason.

PR #84, opened 2024-04-16
fix try_acquire for wasm target

jozanza:
The wasm target's `Client::try_acquire` method doesn't compile because the return value doesn't match the method's signature. This PR fixes that by wrapping the return value in a `Result` which adheres to the method's signature.


weihanglo (approved):
Thanks for the bugfix!

PR #86, opened 2024-04-18
ci: test wasm32-unknown-unknown

weihanglo:
Only test wasm32-unknown-unknown on Ubuntu


weihanglo:
The CI is supposed to fail.

@petrochenkov, to address <https://github.com/rust-lang/jobserver-rs/pull/85>, guess we need this and #84, and release again then.

PR #88, opened 2024-04-20
fix: remove preadv2 optimization

weihanglo:
This optimization has introduced many compatibility issues, so remove it for now until we find it worthy the investment.

See also

* <https://github.com/rust-lang/cc-rs/issues/1039>
* <https://github.com/rust-lang/jobserver-rs/issues/87>

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -285,33 +285,6 @@ impl Client {
     pub fn try_acquire(&self) -> io::Result<Option<Acquired>> {
         let mut buf = [0];
 
-        // On Linux, we can use preadv2 to do non-blocking read,
-        // even if `O_NONBLOCK` is not set.
-        //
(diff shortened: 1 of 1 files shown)


the8472:
You can use `libc::syscall(SYS_preadv2, ...)` to invoke the function without direct glibc support. If the kernel doesn't support it you'll get ENOSYS as usual.

weihanglo:
Thanks! That sounds also correct, though I personally don't have time investing in it (also have less experience of that). Need somebody else to work on that route.

ofek:
If it's not a one or two line change then I agree that we should revert for now to fix builds. I would be interested in the syscall change however because that seems simple but I'm not familiar with the code base.

ofek:
When might this be released?

NobodyXu:
I've asked @petrochenkov to do so on zulip, only they have the right to publish a new release

PR #90, opened 2024-04-22
Add back `preadv2` optimization for `try_acquire` on Linux

### LINKED ISSUE #87: preadv2 not found during linking
Ref: rust-lang/cc-rs#1039

On older glibc version, the preadv2 symbol isn't available.

I think it might makes sense to use syscall directly to avoid this link-time issue.

NobodyXu:
Fixed #87


NobodyXu:
cc @the8472 can you check the syscall I added please?

I'm not sure if it is correct, thank you!

src/unix.rs
```diff
@@ -285,6 +285,33 @@ impl Client {
     pub fn try_acquire(&self) -> io::Result<Option<Acquired>> {
         let mut buf = [0];
 
+        // On Linux, we can use preadv2 to do non-blocking read,
+        // even if `O_NONBLOCK` is not set.
+        //
+        // TODO: musl libc supports preadv2 since 1.2.5, but `libc` crate
+        // hasn't yet added it.
```
petrochenkov, line 292: The TODO is no longer relevant, since bare syscall is used.

src/unix.rs
```diff
@@ -368,6 +395,84 @@ impl Client {
     }
 }
 
+// This should be available for all linux targets,
+// though only [`non_blocking_read`] currently uses it so adding gnu cfg.
```
petrochenkov, line 399: Another outdated comment.

src/unix.rs
```diff
@@ -368,6 +395,84 @@ impl Client {
...
+        iov: *const iovec,
+        iovcnt: c_int,
+        offset: off_t,
+        flags: c_int,
+    ) -> ssize_t {
+        #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
+        let res = syscall(SYS_preadv2, fd, iov, iovcnt, offset, 0, flags);
+
+        #[cfg(all(target_arch = "x86_64", target_pointer_width = "32"))]
+        let res = syscall(SYS_preadv2, fd, iov, iovcnt, offset, 0, flags);
```
petrochenkov, line 427: I don't see any difference between `target_pointer_width` 64 and 32.

src/unix.rs
```diff
@@ -368,6 +395,84 @@ impl Client {
...
 
+// This should be available for all linux targets,
+// though only [`non_blocking_read`] currently uses it so adding gnu cfg.
+#[cfg(any(target_os = "linux", target_os = "android"))]
+mod linux {
+    use super::*;
+
+    use libc::{iovec, off_t, ssize_t, syscall, SYS_preadv2};
+
+    const RWF_NOWAIT: c_int = 0x00000008;
```
the8472, line 406: That seems a little iffy. Theoretically the values could vary between architectures.
Musl added that flag recently. Can you leave a not to use that in the future once libc has it?

the8472 (commented):
Maybe we can simplify things by limiting to aarch64, x86_64 linux gnu? Those are probably the most common environments where stuff is _built_. And it's mostly a perf optimization, so we can be conservative in which platforms get it.

src/unix.rs
```diff
@@ -368,6 +395,84 @@ impl Client {
...
+        }) {
+            Ok(cnt) => Ok(cnt.try_into().unwrap()),
+            Err(err) if err.raw_os_error() == Some(libc::EOPNOTSUPP) => {
+                IS_NONBLOCKING_READ_UNSUPPORTED.store(true, Ordering::Relaxed);
+                Err(io::ErrorKind::Unsupported.into())
+            }
+            Err(err) if err.kind() == io::ErrorKind::Unsupported => {
+                IS_NONBLOCKING_READ_UNSUPPORTED.store(true, Ordering::Relaxed);
+                Err(err)
+            }
```
workingjubilee, line 470: can this just be
```suggestion
            Err(err) if matches!(err.raw_os_error(), Some(libc::EOPNOTSUPP | libc::ENOSYS)) => {
                IS_NONBLOCKING_READ_UNSUPPORTED.store(true, Ordering::Relaxed);
                Err(io::ErrorKind::Unsupported.into())
            }
```

src/unix.rs
```diff
@@ -368,6 +395,84 @@ impl Client {
...
+            fd,
+            iov,
+            iovcnt,
+            offset as libc::c_long,
+            ((offset as u64) >> 32) as libc::c_long,
+            flags,
+        );
+
+        res.try_into().unwrap()
+    }
```
workingjubilee, line 441: While this is a faithful implementation of preadv2, we _could_ simply ditch the args we don't ever use and use the opportunity to comment on why we always pass `iovcnt: 1, offset: -1, flags: RWF_NOWAIT`.

workingjubilee (commented):
what are the numbers on this optimization like?

NobodyXu:
> what are the numbers on this optimization like?

It would enable `try_acquire` support on cases where:
 - jobserver gets an annoymous pipe instead of named fifo
 - procfs is not available
Without `preadv2`, `try_acquire` would return a not-supported error.

On performance, it's mostly for avoiding performance loss with `acquire` when jobserver receives a named fifo or procfs is available (which jobserver would use to convert pipe to fifo).

Without `preadv2`,`try_acquire` would set the fifo to non-blocking, which would cause `acquire` to enter poll + read if the fifo is empty, which is less efficient than a blocking read.

NobodyXu:
> Maybe we can simplify things by limiting to aarch64, x86_64 linux gnu? Those are probably the most common environments where stuff is _built_.

Well, from the glibc implementation, it looks like there're only 3 versions for it, so maybe we can still bare with the cost?

> And it's mostly a perf optimization, so we can be conservative in which platforms get it.

While it's mostly a perf optimization, there is situation where `preadv2` actually enable `try_acquire` to work properly instead of returning a not-supported error:

 - jobserver gets an annoymous pipe instead of named fifo
 - procfs is not available
Without `preadv2`, `try_acquire` would return a not-supported error.

src/unix.rs
NobodyXu: Thanks, I made a mistake when copying the code.

weihanglo:
> > what are the numbers on this optimization like?
> 
> It would enable `try_acquire` support on cases where:
> 
>     * jobserver gets an annoymous pipe instead of named fifo
> 
>     * procfs is not available
>       Without `preadv2`, `try_acquire` would return a not-supported error.
> 
> 
> On performance, it's mostly for avoiding performance loss with `acquire` when jobserver receives a named fifo or procfs is available (which jobserver would use to convert pipe to fifo).
> 
> Without `preadv2`,`try_acquire` would set the fifo to non-blocking, which would cause `acquire` to enter poll + read if the fifo is empty, which is less efficient than a blocking read.

Thanks @NobodyXu for the work. Believing that people already know the benefit of this for fixing unsupported situations. What Jubilee asked for are numbers. I think that's a reasonable request for a change aiming at optimization. Could you do some benchmark and show the number, so we can have a baseline in mind? It is also beneficial to prevent from future performance regression if we have the reproducible benchmark steps (if you don't mind sharing).

src/unix.rs
```diff
@@ -285,6 +285,33 @@ impl Client {
...
+            let read = self.read().as_raw_fd();
+            loop {
+                match linux::non_blocking_read(read, &mut buf) {
+                    Ok(1) => return Ok(Some(Acquired { byte: buf[0] })),
+                    Ok(_) => {
+                        return Err(io::Error::new(
+                            io::ErrorKind::UnexpectedEof,
+                            "early EOF on jobserver pipe",
+                        ))
+                    }
```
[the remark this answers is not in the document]
NobodyXu: It's actually trying to match situations `Ok(0)` here.

I wrote it like this because elsewhere matches the return value of `read` in the same way.

src/unix.rs
NobodyXu: Thanks, updated.

src/unix.rs
NobodyXu: Thanks, removed.

src/unix.rs
NobodyXu: Updated.

src/unix.rs
NobodyXu: Thanks, added a TODO for this.

workingjubilee:
Oh, cute, so it is in fact a kernel optimization specifically for this protocol!

src/unix.rs
```diff
@@ -368,6 +392,83 @@ impl Client {
...
+        if t == -1 {
+            Err(io::Error::last_os_error())
+        } else {
+            Ok(t)
+        }
+    }
+
+    fn preadv2(fd: c_int, iov: &iovec) -> ssize_t {
+        let iovcnt: c_int = 1;
+
```
workingjubilee, line 415: I meant something more like,
```suggestion
        let offset = -1; // "use and update current offset" as in readv, not preadv

```

bit of a magic number otherwise.

src/unix.rs
```diff
@@ -368,6 +392,83 @@ impl Client {
...
+        #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
+        let res = unsafe {
+            syscall(
+                SYS_preadv2,
+                fd,
+                iov,
+                iovcnt,
+                -1 as off_t,
```
workingjubilee, line 423:
```suggestion
                offset as off_t,
```

src/unix.rs
```diff
@@ -368,6 +392,83 @@ impl Client {
...
+        let res = unsafe { syscall(SYS_preadv2, fd, iov, iovcnt, -1 as off_t, RWF_NOWAIT) };
+
+        #[cfg(not(target_arch = "x86_64"))]
+        let res = unsafe {
+            syscall(
+                SYS_preadv2,
+                fd,
+                iov,
+                iovcnt,
+                -1 as libc::c_long,
```
workingjubilee, line 439:
```suggestion
                offset as libc::c_long,
```

src/unix.rs
```diff
@@ -368,6 +392,83 @@ impl Client {
...
+                iov,
+                iovcnt,
+                -1 as off_t,
+                0 as off_t,
+                RWF_NOWAIT,
+            )
+        };
+
+        #[cfg(all(target_arch = "x86_64", target_pointer_width = "32"))]
+        let res = unsafe { syscall(SYS_preadv2, fd, iov, iovcnt, -1 as off_t, RWF_NOWAIT) };
```
workingjubilee, line 430:
```suggestion
        let res = unsafe { syscall(SYS_preadv2, fd, iov, iovcnt, offset as off_t, RWF_NOWAIT) };
```

petrochenkov:
Can the code generally minimize amount of `#[cfg(...)]` in favor of `cfg!(...)`?
It will get more checking and less surprises on remaining platforms.
`#[cfg(...)]` is not really needed that often, only when some *names* are unavailable on some platforms (and even then they can sometimes be defined to dummies, see e.g. how `type RawFd` is defined).

src/unix.rs
NobodyXu: Thanks, I've added the `offset` variable with the right typing.

src/unix.rs
NobodyXu: Updated.

NobodyXu:
> Can the code generally minimize amount of `#[cfg(...)]` in favor of `cfg!(...)`? It will get more checking and less surprises on remaining platforms.

Thanks, I've added the code in `preadv2` to use `cfg!` instead of `#[cfg(...)]`, it's actually more readable now.

src/unix.rs
```diff
@@ -368,6 +392,73 @@ impl Client {
...
+        } else {
+            Ok(t)
+        }
+    }
+
+    fn preadv2(fd: c_int, iov: &iovec) -> ssize_t {
+        let iovcnt: c_int = 1;
+        let offset: off_t = -1;
+
+        if cfg!(all(target_arch = "x86_64", target_pointer_width = "64")) {
```
the8472, line 417: The branches here could use some comment. It's a bit surprising that 86-64 needs two special-cases but all other platforms (including powerpc, aarch64 and even  i686) all go in the same bucket.

the8472 (commented):
Since this code path is trying to achieve something very specific - doing a non-blocking read without setting a pipe fd to O_NONBLOCK - there should be a platform-specific test that checks that this indeed does happen.

But even if we have tests CI only exercises x86-64. If we had targeted tests we could manually run it on the arm cloud machines at least.

My other concern is that the syscall code _looks_ correct but usually the devil is in the details. Having a targeted test would help with that too.

NobodyXu:
Yeah I would add one, I wish libc would support it for musl libc

Or perhaps we could wait till libc support it on musl @the8472 ?

the8472:
Yeah it's not urgent. try_acquire is new and I assume not many things are using it yet so the negative performance impact of making the pipe non-blocking isn't important yet.

NobodyXu:
> try_acquire is new and I assume not many things are using it yet so the negative performance impact of making the pipe non-blocking isn't important yet.

Well cc is using it so it does have some effect to the ecosystem as a whole, though depending on the project, it might not be that large as compiling is much more expensive.

NobodyXu:
Opened rust-lang/libc#3760

PR #98, opened 2024-05-31
Drop msrv to 1.56

udoprog:
MSRV was bumped due to a dev dependency and tests, but dev dependencies and tests are not used by consumers of the library.

This drops MSRV back to the lowest value I could make work without having to hack around too much.

## Using `--remove-dev-deps` with `cargo hack`

Note that this does not affect consumers of this library and is only for local CI.

Note that `cargo hack` needs to use `--remove-dev-deps`. I'm not sure why exactly, but cargo 1.56 seems to misbehave when resolving the `tempfile` dev-dependency.

```
error: failed to select a version for the requirement `rustix = "=0.38.31"`
candidate versions found which didn't match: 0.38.9, 0.38.8, 0.38.7, ...
location searched: crates.io index
required by package `tempfile v3.10.1`
    ... which satisfies dependency `tempfile = "=3.10.1"` of package `jobserver v0.1.31 (D:\Repo\jobserver-rs)`
```

The rustix version *is* available, and I can't figure out why it doesn't think it's a candidate.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -107,4 +107,4 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       - uses: Swatinem/rust-cache@v2
 
-      - run: cargo hack check --lib --rust-version --ignore-private --locked
+      - run: cargo hack check --lib --rust-version --ignore-private --remove-dev-deps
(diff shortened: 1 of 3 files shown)


udoprog:
Ah, there is the `BorrowedFd` use. I don't care enough to have that removed.

Issue #64, opened 2023-12-25
from_env causes io-unsafety in child processes


the8472:
(copied from https://github.com/rust-lang/rust/pull/113730)

These lines seem questionable: https://github.com/rust-lang/jobserver-rs/blob/b4bc5db730a93d01cf8d0d868a2b6f553f535d2d/src/unix.rs#L152-L153

They lead the fds not being available to child processes by default (unless reverted by `Client::configure`) but [from_env_ext()](https://github.com/rust-lang/jobserver-rs/blob/b4bc5db730a93d01cf8d0d868a2b6f553f535d2d/src/lib.rs#L264-L296) does not remove the environment variables. Which means child processes are instructed to access file descriptor numbers for jobserver communication that aren't open anymore and may have been reopened to point to other files. This seems like a violation of IO-safety (https://github.com/rust-lang/rust/issues/116059#issuecomment-1733591651).

Either cloexec shouldn't be set or the environment variables should also be removed and only be added back via `configure`.

NobodyXu:
I think removing the env arg makes sense?

petrochenkov:
I suspect that in a typical case you do want to pass the created or inherited jobserver to children processes.

In all cases caught by https://github.com/rust-lang/rust/pull/113730 the right solution was to pass the jobserver further (except one case when it didn't matter much).

petrochenkov:
https://github.com/rust-lang/jobserver-rs/pull/65 is a draft PR that reverses the current behavior - jobserver is inherited by default, but the inheritance can be disabled by explicitly calling a method.

NobodyXu:
I think the best way to fix this is to use named fifo instead of annoymous pipe.

Passing pipe by fd also shares its file description, meaning whenever one process sets `O_NONBLOCK` it affects all the other processes.

I've got bitten by this the hard way in cc

PR #100, opened 2024-06-29
Fix `Client::configure*` on unix

### LINKED ISSUE #99: jobserver updates jobserver-auth from old style R,W to new style fifo:PATH for children though children may not be compatible with fifo:PATH style auth
Hello,

As discussed in this zulip thread: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/does.20Cargo.20inject.20fifo.3APATH.20style.20jobserver-auth.20in.20MAKEFLAG/near/447522824

We have a build case with rust that started failing on newer nightly toolchains, the crash stemmed from a fifo:PATH style jobserver-auth being forwarded in a build.rs to a make command (with version < 4.4) that is not compatible with the fifo:PATH style auth, when the make command tried to run it crashed indicating it did not support the fifo:PATH auth.

@the8472 pointed out this piece of code https://github.com/rust-lang/jobserver-rs/blob/c0c2898efbdc616c00f86ecfa68831e74488bba3/src/unix.rs#L168-L181

which breaks compatibility if the new fifo:PATH auth is forwarded to incompatible children processes.

Thanks a lot for the swift support!

Cheers

NobodyXu:
Fixed #99

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -6,7 +6,7 @@ use std::io::{self, Read, Write};
 use std::mem;
 use std::mem::MaybeUninit;
 use std::os::unix::prelude::*;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::process::Command;
 use std::ptr;
 use std::sync::{
@@ -17,17 +17,19 @@ use std::thread::{self, Builder, JoinHandle};
 use std::time::Duration;
 
 #[derive(Debug)]
-pub enum Client {
-    /// `--jobserver-auth=R,W`
-    Pipe { read: File, write: File },
-    /// `--jobserver-auth=fifo:PATH`
-    Fifo {
-        file: File,
-        path: PathBuf,
-        /// it can only go from false -> true but not the other way around, since that
-        ///  could cause a race condition.
-        is_non_blocking: AtomicBool,
-    },
+enum ClientCreationArg {
+    Fds { read: c_int, write: c_int },
+    Fifo(Box<Path>),
+}
+
+#[derive(Debug)]
+pub struct Client {
+    read: File,
+    write: File,
+    creation_arg: ClientCreationArg,
+    /// it can only go from Some(false) -> Some(true) but not the other way around, since that
+    ///  could cause a race condition.
+    is_non_blocking: Option<AtomicBool>,
 }
 
 #[derive(Debug)]
@@ -43,7 +45,7 @@ impl Client {
         // wrong!
         const BUFFER: [u8; 128] = [b'|'; 128];
 
-        let mut write = client.write();
+        let mut write = &client.write;
 
         set_nonblocking(write.as_raw_fd(), true)?;
 
@@ -111,16 +113,20 @@ impl Client {
             FromEnvErrorInner::CannotParse("expected a path after `fifo:`".to_string())
         })?;
         let path = Path::new(path_str);
+
         let file = OpenOptions::new()
             .read(true)
             .write(true)
             .open(path)
             .map_err(|err| FromEnvErrorInner::CannotOpenPath(path_str.to_string(), err))?;
 
-        Ok(Some(Client::Fifo {
-            file,
-            path: path.into(),
-            is_non_blocking: AtomicBool::new(false),
+        Ok(Some(Client {
+            read: file
+                .try_clone()
+                .map_err(|err| FromEnvErrorInner::CannotClone(file.as_raw_fd(), err))?,
+            write: file,
+            creation_arg: ClientCreationArg::Fifo(path.into()),
+            is_non_blocking: Some(AtomicBool::new(false)),
         }))
     }
 
@@ -148,6 +154,8 @@ impl Client {
             return Err(FromEnvErrorInner::NegativeFd(write));
         }
 
+        let creation_arg = ClientCreationArg::Fds { read, write };
+
         // Ok so we've got two integers that look like file descriptors, but
         // for extra sanity checking let's see if they actually look like
         // valid files and instances of a pipe if feature enabled before we
@@ -174,40 +182,36 @@ impl Client {
                 //
                 // I tested this on macOS 14 and Linux 6.5.13
                 #[cfg(target_os = "linux")]
-                if let Ok(Some(jobserver)) =
-                    Self::from_fifo(&format!("fifo:/dev/fd/{}", read.as_raw_fd()))
-                {
-                    return Ok(Some(jobserver));
+                if let (Ok(read), Ok(write)) = (
+                    File::open(format!("/dev/fd/{}", read)),
+                    OpenOptions::new()
+                        .write(true)
+                        .open(format!("/dev/fd/{}", write)),
+                ) {
+                    return Ok(Some(Client {
+                        read,
+                        write,
+                        creation_arg,
+                        is_non_blocking: Some(AtomicBool::new(false)),
+                    }));
                 }
             }
         }
 
-        Ok(Some(Client::Pipe {
+        Ok(Some(Client {
             read: clone_fd_and_set_cloexec(read)?,
             write: clone_fd_and_set_cloexec(write)?,
+            creation_arg,
+            is_non_blocking: None,
         }))
     }
 
     unsafe fn from_fds(read: c_int, write: c_int) -> Client {
-        Client::Pipe {
+        Client {
             read: File::from_raw_fd(read),
             write: File::from_raw_fd(write),
-        }
-    }
-
-    /// Gets the read end of our jobserver client.
-    fn read(&self) -> &File {
-        match self {
-            Client::Pipe { read, .. } => read,
-            Client::Fifo { file, .. } => file,
-        }
-    }
-
-    /// Gets the write end of our jobserver client.
-    fn write(&self) -> &File {
-        match self {
-            Client::Pipe { write, .. } => write,
-            Client::Fifo { file, .. } => file,
+            creation_arg: ClientCreationArg::Fds { read, write },
+            is_non_blocking: None,
         }
     }
 
@@ -245,7 +249,7 @@ impl Client {
         // to shut us down, so we otherwise punt all errors upwards.
         unsafe {
             let mut fd: libc::pollfd = mem::zeroed();
-            let mut read = self.read();
+            let mut read = &self.read;
             fd.fd = read.as_raw_fd();
             fd.events = libc::POLLIN;
             loop {
@@ -284,19 +288,15 @@ impl Client {
 
     pub fn try_acquire(&self) -> io::Result<Option<Acquired>> {
         let mut buf = [0];
+        let mut fifo = &self.read;
 
-        let (mut fifo, is_non_blocking) = match self {
-            Self::Fifo {
-                file,
-                is_non_blocking,
-                ..
-            } => (file, is_non_blocking),
-            _ => return Err(io::ErrorKind::Unsupported.into()),
-        };
-
-        if !is_non_blocking.load(Ordering::Relaxed) {
-            set_nonblocking(fifo.as_raw_fd(), true)?;
-            is_non_blocking.store(true, Ordering::Relaxed);
+        if let Some(is_non_blocking) = self.is_non_blocking.as_ref() {
+            if !is_non_blocking.load(Ordering::Relaxed) {
+                set_nonblocking(fifo.as_raw_fd(), true)?;
+                is_non_blocking.store(true, Ordering::Relaxed);
+            }
+        } else {
+            return Err(io::ErrorKind::Unsupported.into());
         }
 
         loop {
@@ -323,7 +323,7 @@ impl Client {
         // always quickly release a token). If that turns out to not be the
         // case we'll get an error anyway!
         let byte = data.map(|d| d.byte).unwrap_or(b'+');
-        match self.write().write(&[byte])? {
+        match (&self.write).write(&[byte])? {
             1 => Ok(()),
             _ => Err(io::Error::new(
                 io::ErrorKind::Other,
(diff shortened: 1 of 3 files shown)


src/unix.rs
```diff
@@ -174,41 +174,42 @@ impl Client {
                 //
                 // I tested this on macOS 14 and Linux 6.5.13
                 #[cfg(target_os = "linux")]
-                if let Ok(Some(jobserver)) =
+                if let Ok(Some(mut jobserver)) =
                     Self::from_fifo(&format!("fifo:/dev/fd/{}", read.as_raw_fd()))
                 {
+                    jobserver.path.take();
```
the8472, line 180: This is fairly indirect in the way it's achieving its goal. At the very least it deserves a comment

src/unix.rs
```diff
@@ -17,17 +17,15 @@ use std::thread::{self, Builder, JoinHandle};
...
-    },
+pub struct Client {
+    read: File,
+    write: Option<File>,
+    /// If path is not None, then a fifo jobserver is created.
+    path: Option<Box<Path>>,
+    supports_non_blocking: bool,
+    /// it can only go from false -> true but not the other way around, since that
+    ///  could cause a race condition.
+    is_non_blocking: AtomicBool,
```
the8472, line 28: The new structure is confusing. `read` sometimes also means write.
The old structure was maybe a little bit more readable, covering different cases explicitly, but it was still a lossy transform of what the parent passed.

Additionally it does not track well whether we have inherited things from a parent or crated our own, which can affect how we should handle environment variables (which could be relevant to #64)

We should do more than fix a bug, we should also keep things organized.

NobodyXu: > Additionally it does not track well whether we have inherited things from a parent or crated our own

This js tracked by the `path` variable, if it is `Some` then we inherited/created a fifo, or we inherited/created a pipe.

NobodyXu: > The new structure is confusing. read sometimes also means write.

I could convert that back to an `enum`

the8472: The thing is we're tracking a bunch of flags and implied state that affects other fields. That obfuscates state.
As I said, the old enum was a bit better, but not good enough either since it did not to cover some state.

I think there are two different aspects
* what we can use ourselves
  * which might be 2 fds, 1 fifo or 2fds+1fifo
  * the blockingness
* what we give to children

NobodyXu: I think maybe we should unify the fifo/pipe?

```rust
struct Client {
    read: File,
    write: File,
    fifo_path: Option<Box<Path>>,
    is_blocking: Option<AtomicBool>,
}
```

It would require an additional open for fifo and an additional fd, but guaranteed it reduces complexity maybe we should adopt it?

It would also separate the read/write, and make sure setting non-blocking on read end would not affect the write end.

the8472: > I think maybe we should unify the fifo/pipe?

I don't think we should do that. In the case where we inherit 2 FDs but reopen it via procfs it would make the unused FD numbers (which we would still pass to children) float around implicitly.
We should encode this explicitly. `std` now has a debug assert that will try to check that owned FDs are not closed twice. So if `jobserver` claims ownership of an FD this will be checked in debug mode.

NobodyXu: No I mean we should open the fifo twice:

```rust
let read = open_read(fifo_path?);
let write = open_write(fifo_path?);

Client {
    read,
    write,
    ...
}
```

Then when we set `O_NONBLOCK` on the read end, the write would be still in blocking mode instead of being changed to non-blocking mode.

the8472: That'd could work for the "real" fifo case but not for the optimization where we inherit 2 fds (and keep passing those to children) and additionally open a process-private one.

NobodyXu: > but not for the optimization where we inherit 2 fds

I think it should work, previously we use `/dev/fd/$fd` to reopen the read fd as readable and writable.

Better solution would be to re-open read end as readable, write end as writable.

I would try this locally and see if it works.

the8472: My point is that we should _also_ keep track of the original FDs because those are the ones we're going to pass to the children.

NobodyXu: Aha right, yes we should do that, because we set the non-blocking on read end.

NobodyXu: So, I propose something like:

```rust
enum ClientSource {
    Fds {
        read: RawFd,
        write: RawFd,
    },
    Fifo(Box<Path>),
}

struct Client {
    read: File,
    write: File,
    is_non_blocking: Option<AtomicBool>,
    source: ClientSource,
}
```

NobodyXu: @the8472 Does that look good to you?

the8472: Hrrm, it looks better. We probably should keep track of the environment too but that's not related to the issue this PR is trying to fix so it can happen elsewhere.

I'll have to see the read/write ends in action to see if it makes sense.

NobodyXu: Updated, turns out for the real fifo, I can open it only once within a process with RW, if I open it with readonly, then open it with write only would block, so I instead clone the fd.

For the optimization of creating a fifo from pipe, it seems to work well.

src/unix.rs
```diff
@@ -17,17 +17,19 @@ use std::thread::{self, Builder, JoinHandle};
...
+}
+
+#[derive(Debug)]
+pub struct Client {
+    read: File,
+    write: File,
+    creation_arg: ClientCreationArg,
+    /// it can only go from Some(false) -> Some(true) but not the other way around, since that
+    ///  could cause a race condition.
+    is_non_blocking: Option<AtomicBool>,
```
NobodyXu, line 32: We could have another

```suggestion
    non_blocking_read: Option<(File, AtomicBool)>,
```

which should improve performance of `acquire`, after `try_acquire` is called, while also enable the fifo optimization for jobserver created by `Client::mk`.

But given how rare `try_acquire` is used, I suspect that is unnecessary, and with #90 it won't be needed on newer Linux kernel anyway.

src/unix.rs
```diff
@@ -17,17 +17,19 @@ use std::thread::{self, Builder, JoinHandle};
...
-    /// `--jobserver-auth=fifo:PATH`
-    Fifo {
-        file: File,
-        path: PathBuf,
-        /// it can only go from false -> true but not the other way around, since that
-        ///  could cause a race condition.
-        is_non_blocking: AtomicBool,
-    },
+enum ClientCreationArg {
+    Fds { read: c_int, write: c_int },
```
NobodyXu, line 21: Was thinking about `BorrowedFd<'static>` here to introduce niche for enum optimization:

```suggestion
    Fds { read: BorrowedFd<'static>, write: BorrowedFd<'static> },
```

But I was afraid that it would introduce more problem, as now fd unsoundness becomes a potentially lifetime unsoundness issue, so I didn't do it.

NobodyXu: ~~Or we could simply store a `Box<str>` here instead, since it is only used in `string_arg()` and returns a `&str` anyway.~~

Not possible as we actually care which variant is used here for `configure`.

src/unix.rs
```diff
@@ -333,31 +333,30 @@ impl Client {
...
     pub fn available(&self) -> io::Result<usize> {
         let mut len = MaybeUninit::<c_int>::uninit();
-        cvt(unsafe { libc::ioctl(self.read().as_raw_fd(), libc::FIONREAD, len.as_mut_ptr()) })?;
+        cvt(unsafe { libc::ioctl(self.read.as_raw_fd(), libc::FIONREAD, len.as_mut_ptr()) })?;
         Ok(unsafe { len.assume_init() } as usize)
     }
 
     pub fn configure(&self, cmd: &mut Command) {
-        match self {
+        if matches!(self.creation_arg, ClientCreationArg::Fifo { .. }) {
```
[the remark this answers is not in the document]
the8472: I think this should happen in a separate PR since we need to consider both the environment and the file descriptors together, basically #64 but also the role of `client.configure`. It'd be a breaking change of that API. Right now the docs say that only children built with this method should inherit the jobserver.

The goal of this PR is to fix the fd to fifo conversion and I think it's already big enough for that.

src/unix.rs
```diff
@@ -284,19 +288,15 @@ impl Client {
...
-
-        if !is_non_blocking.load(Ordering::Relaxed) {
-            set_nonblocking(fifo.as_raw_fd(), true)?;
-            is_non_blocking.store(true, Ordering::Relaxed);
+        if let Some(is_non_blocking) = self.is_non_blocking.as_ref() {
+            if !is_non_blocking.load(Ordering::Relaxed) {
+                set_nonblocking(fifo.as_raw_fd(), true)?;
+                is_non_blocking.store(true, Ordering::Relaxed);
+            }
+        } else {
```
the8472, line 298: the check isn't quite correct since an inherited FD from make may already be in nonblocking mode. We can't modify the flag when sharing the file description with other processes, but we can still opportunistically use it.

Which also means the `is_non_blocking` name is misleading.

src/unix.rs
```diff
@@ -111,16 +113,20 @@ impl Client {
...
             .map_err(|err| FromEnvErrorInner::CannotOpenPath(path_str.to_string(), err))?;
 
-        Ok(Some(Client::Fifo {
-            file,
-            path: path.into(),
-            is_non_blocking: AtomicBool::new(false),
+        Ok(Some(Client {
+            read: file
+                .try_clone()
+                .map_err(|err| FromEnvErrorInner::CannotClone(file.as_raw_fd(), err))?,
```
the8472, line 126: Introducing a public new error case just because we internally happen to clone an FD to keep two fields populated even though it's not strictly necessary does not seem like a good tradeoff.

In that case having another enum is probably better.

src/unix.rs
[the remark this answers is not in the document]
NobodyXu: Got it thanks

src/unix.rs
NobodyXu: > but we can still opportunistically use it.

We can't do it, since it is shared between processes, other processes can clear the non-blocking flags and convert it to blocking at anytime.

It can happen after we checked that it has non-blocking set and start a non-blocking read, or it can happen during the read, which I'm not sure how the kernel will handle it but it's pretty bad.

src/unix.rs
NobodyXu: I could reuse `CannotOpenFd`, does that look good to you?

src/unix.rs
the8472: > Updated, turns out for the real fifo, I can open it only once within a process with RW

Wouldn't that be a problem for the real fifo case? `Client::from_env` docs say

> It is ok to call this function any number of times.

If this would block that could cause issues for the caller.

Though I'm not seeing the "open twice" scenario mentioned in `man 7 fifo`. Maybe `open(..., O_NONBLOCK)` and then unsetting it afterwards via `fcntl` would help?

src/unix.rs
the8472: Other processes shouldn't be manipulating that flag on a shared resource. But yeah, I guess it's better to be on the safe side.

src/unix.rs
the8472: Good enough

the8472:
This also needs a regression test. Apparently nothing ensured that when inheriting a configuration from the parent we pass it through properly to a child.

src/unix.rs
NobodyXu: > Wouldn't that be a problem for the real fifo case? Client::from_env docs say

In the test, the `Client` is the only one opening the fifo.

https://stackoverflow.com/questions/76411975/is-there-a-way-to-open-fifonamed-pipe-in-write-only-way-without-blocked

> From [the FIFO manual page](https://man7.org/linux/man-pages/man7/fifo.7.html):
>
> > Normally, opening the FIFO blocks until the other end is opened also.

So I guess open it with read + write makes sense and prevent that blocking.

NobodyXu: > Maybe open(..., O_NONBLOCK) and then unsetting it afterwards via fcntl would help?

Yeah that should help:

> A process can open a FIFO in nonblocking mode.  In this case,
       opening for read-only succeeds even if no one has opened on the
       write side yet and opening for write-only fails with ENXIO (no
       such device or address) unless the other end has already been
       opened.

the8472: > We can open the fifo again with read/write only.

Oh ok, then I misunderstood what you wrote.

In that case we could open it twice instead of cloning it, that way we could even manage the O_NONBLOCK flags separately.

src/unix.rs
NobodyXu: Reverted this change.

src/unix.rs
NobodyXu: Thanks, changed it to open it twice.

NobodyXu:
cc @the8472 Updated, added regression test as suggested.

src/unix.rs
```diff
@@ -17,17 +17,19 @@ use std::thread::{self, Builder, JoinHandle};
...
-    Pipe { read: File, write: File },
-    /// `--jobserver-auth=fifo:PATH`
-    Fifo {
-        file: File,
-        path: PathBuf,
-        /// it can only go from false -> true but not the other way around, since that
-        ///  could cause a race condition.
-        is_non_blocking: AtomicBool,
-    },
+enum ClientCreationArg {
```
weihanglo, line 20: There are some complications around this enum. Could we document it? Something like


```suggestion
/// This preserves the `--jobserver-auth` type at creation time,
/// so auth type will be passed down to and inherit from sub-Make processes correctly.
///
/// See <https://github.com/rust-lang/jobserver-rs/issues/99> for details.
enum ClientCreationArg {
```

weihanglo (commented):
Thanks for the fix. Could we fix up some commits?

Also, I'd prefer presenting the regression test first in a commit, followed by the other commits actually fix the behavior and tests. That would make the first commit a minimal reproduction and we know we're fixing the right thing.

src/unix.rs
the8472, line 32: The combination of the name and the `Option` here is not quite obvious. The option indicates whether it **can be** made non-blocking, the boolean indicates whether it **is** nonblocking. The comment could explain that.
It can be understood by reading the code below but having to jump around less should make things easier for new readers.

src/unix.rs
```diff
@@ -111,16 +113,20 @@ impl Client {
...
+            OpenOptions::new()
+                .read(true)
+                .write(true)
+                .open(path)
+                .map_err(|err| FromEnvErrorInner::CannotOpenPath(path_str.to_string(), err))
+        };
+
+        Ok(Some(Client {
+            read: open_file()?,
+            write: open_file()?,
```
the8472, line 127: Opening "read" and "write" as both read-write is a little bit weird. This could be worth 1-2 lines to avoid confusion later.

the8472 (approved):
Functionally it seems fine to me now. I think it could use a few more comments here and there.

NobodyXu:
> Also, I'd prefer presenting the regression test first in a commit, followed by the other commits actually fix the behavior and tests. That would make the first commit a minimal reproduction and we know we're fixing the right thing.

That might be a bit hard since I wrote after fixing it

src/unix.rs
NobodyXu: Thanks, updated

src/unix.rs
NobodyXu: Thanks updated

NobodyXu:
@weihanglo I squashed the commits, can't get the regression test commit to be the first, but it's much smaller now.

src/unix.rs
```diff
@@ -17,17 +17,26 @@ use std::thread::{self, Builder, JoinHandle};
...
+#[derive(Debug)]
+pub struct Client {
+    read: File,
+    write: File,
+    creation_arg: ClientCreationArg,
+    /// It is set to `None` if the pipe is shared with other processesdoes so it cannot support
+    ///  non-blocking mode.
+    ///
+    /// If it is set to `Some`, then It can only go from `Some(false)` -> `Some(true)` but not
+    ///  the other way around, since that could cause a race condition.
```
weihanglo, line 38:
```suggestion
    /// It is set to `None` if the pipe is shared with other processes so it cannot support
    /// non-blocking mode.
    ///
    /// If it is set to `Some`, then it can only go from `Some(false)` -> `Some(true)` but not
    /// the other way around, since that could cause a race condition.
```

src/unix.rs
```diff
@@ -111,16 +120,23 @@ impl Client {
...
-
-        Ok(Some(Client::Fifo {
-            file,
-            path: path.into(),
-            is_non_blocking: AtomicBool::new(false),
+
+        let open_file = || {
+            // Open with read write is necessary, since open with read-only or
+            // write-only could block the thread until another thread open it
+            // with write-only or read-only (or RDWR) correspondingly.
```
weihanglo, line 127:
```suggestion
            // Open with read write is necessary, since opening with read-only or
            // write-only could block the thread until another thread opens it
            // with write-only or read-only (or RDWR) correspondingly.
```

weihanglo (approved):
Typos. Otherwise LGTM

weihanglo:
@petrochenkov 

Would appreciate if you have time publishing this :)

(We should probably revisit <https://github.com/rust-lang/jobserver-rs/pull/89> and <https://github.com/rust-lang/infra-team/issues/117>)

PR #101, opened 2024-07-21
Fix IO safety in `unix.rs`

beetrees:
As well as using `BorrowedFd` as opposed to `c_int` in more places, this PR fixes two related bugs:

- Unix `Client::configure()` wouldn't ensure the file descriptors passed to `Command::configure` remained valid until the command was actually spawned. This meant that if the `Client` was dropped before the `Command` was spawned, the `set_cloexec` call could end up hitting an unrelated file descriptor (which the spawned process would then treat as a jobserver).
- Unix `string_arg` uses the FDs from `ClientCreationArg::Fds` to put in the environment variables, but before this PR `Client::configure` would use the FDs from `self.read`/`self.write` instead. This would result in the wrong FDs getting passed to `set_cloexec`. This was basically harmless (apart from having the use-after-drop problem) as the only time the two sets of FDs differ is when the jobserver has been inherited from the environment, in which case the FDs are going to be not-CLOEXEC anyway.

I've also removed the `unsafe` from the definitions of `Client::mk` and `Client::from_fds` as they don't appear to have any safety requirements.

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -47,7 +47,7 @@ pub struct Acquired {
 
 impl Client {
     pub fn new(mut limit: usize) -> io::Result<Client> {
-        let client = unsafe { Client::mk()? };
+        let client = Client::mk()?;
 
         // I don't think the character written here matters, but I could be
         // wrong!
@@ -55,7 +55,7 @@ impl Client {
 
         let mut write = &client.write;
 
-        set_nonblocking(write.as_raw_fd(), true)?;
+        set_nonblocking(write.as_fd(), true)?;
 
         while limit > 0 {
             let n = limit.min(BUFFER.len());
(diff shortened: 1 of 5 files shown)


src/unix.rs
```diff
@@ -64,39 +64,52 @@ impl Client {
...
-                            PIPE2_AVAILABLE.store(false, Ordering::SeqCst);
-                        } else {
-                            return Err(err);
+                unsafe {
+                    let mut pipes = [0; 2];
+                    match libc::syscall(libc::SYS_pipe2, pipes.as_mut_ptr(), libc::O_CLOEXEC) {
+                        -1 => {
+                            let err = io::Error::last_os_error();
+                            if err.raw_os_error() == Some(libc::ENOSYS) {
+                                PIPE2_AVAILABLE.store(false, Ordering::SeqCst);
```
NobodyXu, line 86: I think `Relaxed` here is definitely enough for store and load

```suggestion
                                PIPE2_AVAILABLE.store(false, Ordering::Relaxed);
```

src/unix.rs
```diff
@@ -64,39 +64,52 @@ impl Client {
...
-                match libc::syscall(libc::SYS_pipe2, pipes.as_mut_ptr(), libc::O_CLOEXEC) {
-                    -1 => {
-                        let err = io::Error::last_os_error();
-                        if err.raw_os_error() == Some(libc::ENOSYS) {
-                            PIPE2_AVAILABLE.store(false, Ordering::SeqCst);
-                        } else {
-                            return Err(err);
+                unsafe {
+                    let mut pipes = [0; 2];
+                    match libc::syscall(libc::SYS_pipe2, pipes.as_mut_ptr(), libc::O_CLOEXEC) {
```
NobodyXu, line 82: Could use `cvt` here and match on the error:

```suggestion
                    match cvt(libc::syscall(libc::SYS_pipe2, pipes.as_mut_ptr(), libc::O_CLOEXEC)) {
```

src/unix.rs
```diff
@@ -64,39 +64,52 @@ impl Client {
...
+        let (read, write) = unsafe {
+            let mut pipes = [0; 2];
+            cvt(libc::pipe(pipes.as_mut_ptr()))?;
+            (
+                OwnedFd::from_raw_fd(pipes[0]),
+                OwnedFd::from_raw_fd(pipes[1]),
+            )
+        };
+        drop(set_cloexec(read.as_fd(), true));
+        drop(set_cloexec(write.as_fd(), true));
```
NobodyXu, line 111: I think the error should be propagated instead of ignored:

```suggestion
        set_cloexec(read.as_fd(), true)?;
        set_cloexec(write.as_fd(), true)?;
```

src/unix.rs
```diff
@@ -64,39 +64,52 @@ impl Client {
...
             static PIPE2_AVAILABLE: AtomicBool = AtomicBool::new(true);
             if PIPE2_AVAILABLE.load(Ordering::SeqCst) {
-                match libc::syscall(libc::SYS_pipe2, pipes.as_mut_ptr(), libc::O_CLOEXEC) {
-                    -1 => {
-                        let err = io::Error::last_os_error();
-                        if err.raw_os_error() == Some(libc::ENOSYS) {
-                            PIPE2_AVAILABLE.store(false, Ordering::SeqCst);
-                        } else {
-                            return Err(err);
+                unsafe {
```
NobodyXu, line 80: My two cent is that smaller `unsafe` block is better, as it makes it easier to read

tests/spawn-after-drop.rs
```diff
@@ -0,0 +1,52 @@
+use std::{env, process::Command};
+
+use jobserver::{Client, FromEnvErrorKind};
+
+macro_rules! t {
```
NobodyXu, line 5: I don't think we need `t!` anymore, it's probably used in times where `.unwrap()` isn't available

src/unix.rs
```diff
@@ -357,24 +373,30 @@ impl Client {
...
+            ClientCreationArg::Fds { read, write } => {
+                // Here we basically just want to say that in the child process
+                // we'll configure the read/write file descriptors to *not* be
+                // cloexec, so they're inherited across the exec and specified as
+                // integers through `string_arg` above.
+                unsafe {
+                    // Keep a reference to the jobserver alive in the closure so that
+                    // the pipe FDs aren't closed, otherwise `set_cloexec` might end up
+                    // targetting a completely unrelated file descriptor.
+                    let arc = self.clone();
```
NobodyXu, line 391: This is awesome, now it would always work once `Client::configure` is called

tests/spawn-after-drop.rs
```diff
@@ -0,0 +1,52 @@
...
+            let status = t!(cmd.arg("from_env").status());
+            assert!(status.success());
+        }
+        "from_env" => {
+            let me = t!(env::current_exe());
+            let mut cmd = Command::new(me);
+            let client = unsafe {
+                t!(match Client::from_env_ext(true).client {
+                    // Its ok for a dropped jobservers path to no longer exist (e.g. on Windows).
+                    Err(e) if matches!(e.kind(), FromEnvErrorKind::CannotOpenPath) => return,
```
NobodyXu, line 31: IMHO it's kind of a bad behavior that it has different behavior on different platform.

NobodyXu (changes requested):

NobodyXu:
>     * On Windows: The command spawns successfully but attempting to inherit the jobserver will error.

That's unfortunate since windows don't have `pre_exec`.

I have a fork of jobserver, which I workaround this issue by forcing user to pass in a closure and spawn the process there [`Client::configure_and_run`](https://docs.rs/jobslot/latest/jobslot/struct.Client.html#method.configure_and_run):

```rust
pub fn configure_and_run<Cmd, F, R>(&self, cmd: Cmd, f: F) -> Result<R>
where
    Cmd: Command,
    F: FnOnce(&mut Cmd) -> Result<R>;
```

>     * On Unix after this PR: The command spawns successfully and the jobserver is inheritable and usable.

I was planning to open a PR to optimize away the `pre_exec`, if the `Client` is created `from_env`, since the original fds inherited from the environment should be valid.

Though that'd mean the jobserver would be inherited without `Client::configure`

tests/spawn-after-drop.rs
michaelvanstraten: @ChrisDenton, one more case to start supporting this.

michaelvanstraten: I am happy to collaborate would @beetrees on this to stop processes from inheriting handles.

michaelvanstraten: Check out https://github.com/rust-lang/rust/pull/115501.

PR #106, opened 2024-09-16
Declare Windows imports jobserver depends on

ChrisDenton:
Jobserver is specifying external functions without also declaring a dependency on the relevant import library that contains those functions. While `kernel32` will often be linked by *something*, the same can not be said for `advapi32`. As a bonus this is mildly more efficient.

In the future we could use `raw-dylib` to avoid the need for import libs entirely (requires msrv to be 1.65+),

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -36,6 +36,7 @@ const WAIT_FAILED: DWORD = 4294967295u32;
 const WAIT_OBJECT_0: DWORD = 0u32;
 const WAIT_TIMEOUT: DWORD = 258u32;
 
+#[link(name = "kernel32")]
 extern "system" {
     fn CloseHandle(handle: HANDLE) -> BOOL;
     fn SetEvent(hEvent: HANDLE) -> BOOL;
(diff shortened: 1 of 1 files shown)


src/windows.rs
weihanglo, line 1: Thanks for the patch!

>  As a bonus this is mildly more efficient.

Out of curiosity, how does it get more efficient with this change?

weihanglo (approved):

src/windows.rs
ChrisDenton: I wrote a bit about something similar on zulip: https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Link.20attribute.20on.20Windows/near/470532686

It's a super minor thing and almost certainly doesn't affect the performance of jobserver. In short, an `extern` block defaults to static linkage whereas `#[link]` defaults to dynamic linkage. Since this functions is from a DLL we want dynamic linkage because the location of a function is stored at an address in the import table rather than a static function you can call directly. 

Static linkage wouldn't work at all except import libraries add a "jump stub". That is a kinda of dummy function that just jumps to the real function. It's the equivalent to this:

```rust
fn CloseHandle(handle: HANDLE) -> BOOL {
    (*__imp_CloseHandle)(handle)
}
```

So there's an extra level of indirection. Like, I said it's super minor and only rarely matters.

PR #110, opened 2025-03-26
Use getrandom for randomness

ChrisDenton:
The reason `getrandom` isn't currently used is given by: https://github.com/rust-lang/jobserver-rs/blob/f494d826e436a7027f40affa504cb8777923d108/src/windows.rs#L71-L76 which says `getrandom` can't be used due to https://github.com/rust-lang/rust/issues/65014 which was closed as an instance of https://github.com/rust-lang/cargo/issues/4866 which is closed in favour of a nightly cargo feature https://github.com/rust-lang/cargo/issues/7915 which has since been stabilised.

So I think we're good now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,9 +37,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25cbce373ec4653f1a01a31e8a5e5ec0c622dc27ff9c4e6606eefef5cbbed4a5"
 
 [[package]]
+name = "getrandom"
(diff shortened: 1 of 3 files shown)


clubby789:
@weihanglo Would it be possible to get a new release of `jobserver` with this? It should unblock [this dependency bump](https://github.com/rust-lang/rust/pull/138848#issuecomment-2753507986)

weihanglo:
[jobserver@0.1.33](https://crates.io/crates/jobserver/0.1.33) is out.

PR #113, opened 2025-08-22
Fix running inside miri when using in-process accounting

bjorn3:
Miri doesn't currently support mkfifo, so can't present named pipes for other processes (but spawning processes isn't possible from miri anyway) And it also isn't possible in miri to get fd's inherited from the parent process. But this is enough for what running rustc in miri needs.

Fixes https://github.com/rust-lang/jobserver-rs/issues/108

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -8,10 +8,9 @@ use std::mem::MaybeUninit;
 use std::os::unix::prelude::*;
 use std::path::Path;
 use std::process::Command;
-use std::ptr;
 use std::sync::{
     atomic::{AtomicBool, Ordering},
-    Arc, Once,
+    Arc,
 };
 use std::thread::{self, Builder, JoinHandle};
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
This seems reasonable.  Thanks.

---

A bit more history:

`pipe2()` was added in kernel 2.6.27 and glibc 2.9. [^0]

Since Rust 1.64.0 (2022-09-22) [^1], we requires

* glibc >= 2.17 (previously glibc >= 2.11)
* kernel >= 3.2 (previously kernel >= 2.6.32)

The oldest version requiring glibc-2.11 / kernel-2.6.32 I can found from doc [^2] is 1.47.0 (2020-10-08). However, our MSRV is already [1.63 one year ago [^3]. If people using toolchain older than 1.47, their build are already broken.

Anyway, it is reasonable. Thanks!


[^0]: https://linux.die.net/man/2/pipe2
[^1]: https://blog.rust-lang.org/2022/08/01/Increasing-glibc-kernel-requirements/
[^2]: https://doc.rust-lang.org/1.47.0/rustc/platform-support.html#tier-1
[^3]: https://crates.io/crates/jobserver/0.1.28

weihanglo:
@bjorn3 jobserver@0.1.34 is out <https://crates.io/crates/jobserver/0.1.34>.

bjorn3:
Thanks!

PR #115, opened 2025-09-19
Fix atomicity violation in try_acquire by using compare_exchange

quantpoet:
Replaced the racy load-then-store pattern with atomic compare_exchange to eliminate  the potential race condition.

This ensures the non-blocking flag is set atomically, preventing multiple threads from simultaneously calling set_nonblocking on the same file descriptor.


tgross35:
> This ensures the non-blocking flag is set atomically, preventing multiple threads from simultaneously calling set_nonblocking on the same file descriptor.

Why is this a problem? It looks like an idempotent fcntl call. Also, wouldn't `swap` be sufficient?

tgross35:
Worse actually: with this PR, thread A can swap `is_non_blocking` from `false` to `true` and start `fnctl`. Thread B may observe `is_non_blocking = true` before thread A's syscall completes and assume the FIFO is `O_NONBLOCK`, even though that isn't correct.

The current implementation ensures `O_NONBLOCK` is set before continuing.

quantpoet:
> Worse actually: with this PR, thread A can swap `is_non_blocking` from `false` to `true` and start `fnctl`. Thread B may observe `is_non_blocking = true` before thread A's syscall completes and assume the FIFO is `O_NONBLOCK`, even though that isn't correct.
> 
> The current implementation ensures `O_NONBLOCK` is set before continuing.

Oh indeed. Thanks for your reply.

PR #117, opened 2025-11-17
jobserver fd_check: Allow sockets.

xim:
I agree with the creator of issue #6 that jobserver fds should be allowed to be sockets as well as pipes. A project I'm working on uses a jobserver implemented over Unix domain sockets.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -32,8 +32,12 @@ pub enum FromEnvErrorKind {
     /// is negative, which means it is disabled for this process
     /// ([GNU `make` manual: POSIX Jobserver Interaction](https://www.gnu.org/software/make/manual/make.html#POSIX-Jobserver)).
     NegativeFd,
-    /// File descriptor from the jobserver environment variable value is not a pipe.
+    /// Deprecated: File descriptor from is not a pipe, specifically.
+    #[deprecated(note = "Use `UnsupportedFdType` instead")]
     NotAPipe,
+    /// File descriptor from the jobserver environment variable value is not a
+    /// supported file type (pipe, socket or character device).
+    UnsupportedFdType,
     /// Jobserver inheritance is not supported on this platform.
     Unsupported,
 }
@@ -48,7 +52,7 @@ impl FromEnvError {
             FromEnvErrorInner::CannotOpenPath(..) => FromEnvErrorKind::CannotOpenPath,
             FromEnvErrorInner::CannotOpenFd(..) => FromEnvErrorKind::CannotOpenFd,
             FromEnvErrorInner::NegativeFd(..) => FromEnvErrorKind::NegativeFd,
-            FromEnvErrorInner::NotAPipe(..) => FromEnvErrorKind::NotAPipe,
+            FromEnvErrorInner::UnsupportedFdType(..) => FromEnvErrorKind::UnsupportedFdType,
             FromEnvErrorInner::Unsupported => FromEnvErrorKind::Unsupported,
         }
     }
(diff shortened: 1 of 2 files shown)


xim:
Can add some details ; I already have a piece of software running that creates a token pool across all builds on the local machine using an abstract unix domain socket. It works well with GNU Make, Ninja and gcc.

xim:
I'm more than willing to contribute, and to consider alternative approaches.

src/error.rs
```diff
@@ -48,7 +48,7 @@ impl FromEnvError {
             FromEnvErrorInner::CannotOpenPath(..) => FromEnvErrorKind::CannotOpenPath,
             FromEnvErrorInner::CannotOpenFd(..) => FromEnvErrorKind::CannotOpenFd,
             FromEnvErrorInner::NegativeFd(..) => FromEnvErrorKind::NegativeFd,
-            FromEnvErrorInner::NotAPipe(..) => FromEnvErrorKind::NotAPipe,
+            FromEnvErrorInner::NotAPipeOrSocket(..) => FromEnvErrorKind::NotAPipeOrSocket,
```
NobodyXu, line 51: This would probably be an undesired breaking change, don't think we can rename this enum variant which is unfortunate

Could add a `NotASocket`

NobodyXu (commented):
I'm not a maintainer of jobserver though I did have some contributions and subscribe to updates to jobserver closely and is a maintainer of cc, which use jobserver, so take my comment with a grain of salt.

And I'm not sure if I should merge the PR here on behalf of the maintainer, so probably not unless nobody has time to review it

src/unix.rs
```diff
@@ -495,7 +495,14 @@ unsafe fn fd_check(fd: c_int, check_pipe: bool) -> Result<(), FromEnvErrorInner>
...
                 return Ok(());
             }
-            Err(FromEnvErrorInner::NotAPipe(fd, None))
+            // Also allow sockets (S_IFSOCK)
+            #[allow(unused_assignments)]
+            let mut s_ifsock = stat.st_mode;
+            s_ifsock = libc::S_IFSOCK as _;
+            if stat.st_mode & s_ifsock == s_ifsock {
+                return Ok(());
+            }
```
CertainLach: There are jobserver implementations that use CUSE (character device in userspace, S_IFCHR), e.g

https://gitweb.gentoo.org/proj/steve.git
https://codeberg.org/amonakov/guildmaster
https://github.com/NixOS/nixpkgs/pull/314888

Would be great if this PR permitted them too:

```suggestion
            let fmt = |v| {
                #[allow(unused_assignments)]
                let mut s_ifmt = stat.st_mode;
                s_ifmt = v as _;
                s_ifmt
            };
            
            let fmt = stat.st_mode & fmt(libc::S_IFMT);
            if fmt == fmt(libc::S_IFIFO) || fmt == fmt(libc::S_IFSOCK) || fmt == fmt(libc::S_IFCHR)  {
                return Ok(());
            }
            Err(FromEnvErrorInner::NotAPipe(fd, None))
```

CertainLach (changes requested):

src/error.rs
xim: I checked all uses in cargo/rust, this value was never explicitly referenced.

src/unix.rs
xim: I did that, with some minor changes. Added you as co-author. Also renamed the error enum and text to be more generic.

src/unix.rs
weihanglo, line 1: Is it possible we have tests for the new behavior including successful paths and failing paths?

src/error.rs
weihanglo: It is better to avoid MSRV bump. I would suggest adding new variant and leave the old one there. We already have `#[non_exhaustive]` so should be fine adding new variants.

xim: Kept the old public variant and marked it as deprecated

src/unix.rs
xim: Sure!

weihanglo (commented):
Thanks for the update. Would you mind rebasing onto master so the CI should pass?

PR #120, opened 2026-06-17
Update getrandom to 0.4

kpcyrd:
I noticed there's a few old crates in my dependency tree, that I traced to `getrandom 0.3`, which was pulled in by `jobserver`, which was pulled in by the `cc` crate.


kpcyrd:
I looked at the errors and it seems to be two root causes:

- GNU make 4.4.1 uses a [known bugged](https://bugs.gentoo.org/945968) gnulib version regarding C23 compatibility
- The 1.63.0 MSRV is currently the Rust version in Debian old-stable, while Debian stable has 1.85.0 (which is also the MSRV of getrandom 0.4)

Cargo.toml
```diff
@@ -19,7 +19,7 @@ libc = "0.2.171"
 nix = { version = "0.28.0", features = ["fs"] }
 
 [target.'cfg(windows)'.dependencies]
-getrandom = { version = "0.3.2", features = ["std"] }
+getrandom = { version = "0.4", features = ["std"] }
```
weihanglo, line 22: CI is failing because getrandom@0.4 requires higher MSRV than jobserver. I think we should check our [reverse-dependencies](https://crates.io/crates/jobserver/reverse_dependencies) and see what is the common MSRV they have today in order to avoid some churns. That said, with Cargo's MSRV resolver, this should not be an issue in general.

Would you mind help do a small analysis?

kpcyrd: - `bizinikiwi-wasm-builder` - 2021 Edition, last release 2025
- `build-parallel` - 2018 Edition, last release 2021
- `cargo-mutants` - **MSRV 1.88.0**
- `cargo-semverver` - 2018 Edition, last release 2019
- `cargo-update` - **MSRV 1.86.0**
- `cargo-util` - **MSRV 1.94.0**
- `cargo` - **MSRV 1.94.0**
- `cc` - **MSRV 1.63.0**
- `cjam` - 2024 Edition, last release 2025
- `fast-rustc-ap-rustc_data_structures` - 2018 Edition, last release 2020
- `fortranc` - 2021 Edition, last release 2025
- `hal` - 2018 Edition, last release 2025
- `hkalbasi-rustc-ap-rustc_data_structures` - 2021 Edition, last release 2023
- `jam-pvm-builder` - 2021 Edition, last release 2026
- `libwild` - **MSRV 1.94.0**
- `msiz_rustc-ap-rustc_data_structures` - 2018 Edition, last release 2019
- `nasm-rs` - 2018 Edition, last release 2026
- `opencv` - **MSRV 1.82.0**
- `pierport` - **MSRV 1.76.0**
- `pyth-price-store` - 2021 Edition, last release 2024
- `ra-ap-rustc_data_structures` - 2021 Edition, last release 2024
- `rstash` - **MSRV 1.75.0**
- `rustc-ap-rustc_data_structures` - 2018 Edition, last release 2021
- `rustc_data_structures` - 2021 Edition, last release 2024
- `sccache` - **MSRV 1.85.0**
- `substrate-wasm-builder` - 2021 Edition, last release 2026
- `tcl-sys` - 2021 Edition, last release 2023
- `tesseract-ocr-static-c` - 2024 Edition, last release 2026
- `vkcargo` - 2018 Edition, last release 2020

weihanglo: Thanks for the analysis!

Looks like cc-rs has a relatively low MSRV, and they are some old discussions around that: <https://github.com/rust-lang/cc-rs/pull/1513>. While this package doesn't get as many updates as cc-rs, I am afraid we might need to stick with it for a while.

weihanglo: @cuviper and @NobodyXu any thoughts? 1.85 is one and a half year old and MSRV aware resolver is stabilized in 1.84, plus jobserver doesn't have that much updates. This should be generally fine?

cuviper: For my part in that discussion (Aug 2025):
> RHEL 8/9/10 only have 1.84 right now, but they'll be upgraded in the next minor release, around November.

We did indeed ship upgrades to RHEL, with 1.88 in November and 1.92 last month.

konstin: As a counterpoint, I found this issue looking if we can remove dependency duplication cause we currently build three versions of getrandom, and jobserver is one of the crates that pulls in an old getrandom version into our dep tree.

djc: Additionally, getrandom 0.3.4 has an unconditional dependency on wasip2 which is dragging in a lot of extra dependencies. I really think we need to bite the bullet here and take the MSRV bump, which should be fine especially for downstreams that can use MSRV-aware resolution.

weihanglo: Makes sense. Thanks folks. Will merge and perhaps cut a new release soon.

weihanglo:
0.1.35 is out https://crates.io/crates/jobserver/0.1.35

